PR #1, opened 2024-03-01
Initial Release

dean-krueger:


src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+    CombineInventory(tritium_storage);
+  }
+
+  // This pulls out some of the blanket each timestep so that fresh blanket can
+  // be added.
+  if (!blanket.empty() &&
+      blanket.quantity() >= blanket_size * blanket_turnover_rate) {
+    cyclus::Material::Ptr blanket_mat = blanket.Pop();
+    cyclus::Material::Ptr spent_blanket =
+        blanket_mat->ExtractQty(blanket_size * blanket_turnover_rate);
```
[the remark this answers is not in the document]
NybergWISC: How easy is it in cyclus to add and remove specific isotopes from the mixture? Tritium may leak/bubble off/be processed out and pure lithium added at different timesteps/schedules

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+            std::to_string(blanket_size * blanket_turnover_rate) + "kg!");
+  }
+}
+
+void Reactor::Tock() {
+  if (!core_loaded) {
+    try {
+      Startup();
+      fuel_startup_policy.Stop();
+      // blanket_startup_policy.Stop();
```
NybergWISC, line 95: Remove commented testing line

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void Reactor::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_startup_policy
+      .Init(this, &tritium_reserve, std::string("Tritium Storage"),
+            &fuel_tracker, std::string("ss"),
+            reserve_inventory + startup_inventory,
+            reserve_inventory + startup_inventory)
```
NybergWISC, line 119: Does this line make the startup requirement equal to reserve+startup inventories? I haven't looked through the docs as much as you, but I thought the startup inventory would count towards the reserve inventory/the reserve inventory is included in the startup inventory

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+      .Set(fuel_incommod)
+      .Start();
+  blanket_fill_policy
+      .Init(this, &blanket, std::string("Blanket Startup"), &blanket_tracker,
+            std::string("ss"), blanket_size, blanket_size)
+      .Set(blanket_incommod)
+      .Start();
+
+  // Tritium Buy Policy Selection:
```
NybergWISC, line 128: For @gonuke, is this behavior reasonable? We thought this was a good simple set of behaviors but it seems like the question of refueling schedules/behavior has probably been addressed elsewhere in cyclus

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+        std::string("Expected Composition: {{10030000,1.000000}}. ") +
+        std::string("Fuel Incommod Composition: ") +
+        std::string(GetComp(initial_reserve)));
+  } else {
+    tritium_reserve.Push(initial_reserve);
+    throw cyclus::ValueError(
+        "Startup Failed: " + std::to_string(tritium_reserve.quantity()) +
+        " kg in reserve is less than required " +
+        std::to_string(startup_inventory + reserve_inventory) +
+        " kg to start-up!");
```
NybergWISC, line 204: Again, my definitions may be off but I thought the amount required for startup was startup_inventory (which includes enough to have a reserve). It's a small difference but may be important so that the model lines up with previous models/variable expectations

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void Reactor::DepleteBlanket(double bred_tritium_mass) {
+  cyclus::Material::Ptr blanket_mat = blanket.Pop();
+
+  cyclus::CompMap b = blanket_mat->comp()->mass();
+  cyclus::compmath::Normalize(&b, blanket_mat->quantity());
+
+  // Percent (as decimal) of T which comes from Li-7 instead of Li-6
+  double Li7_contribution = 0.15;
```
NybergWISC, line 302: On second thought, I think we should make this a modifiable variable even with the simple model. 0.15 seems like a good default that lines up with my simple models (although PW/BL/TB can have input and I can ask Hitarth what he has found)

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+
+  cyclus::CompMap depleted_comp;
+
+  // This is ALMOST the correct behavior, but "scraping the bottom of the
+  // barrel" is a little too complex for this implementation.
+  if ((b[30060000] - (1 - Li7_contribution) * 2 * bred_tritium_mass > 0) &&
```
NybergWISC, line 308: How should we treat the edge case (probably just a test case) of 100% Li-6 or Li7 (Li7= 1 or 0)? Maybe the && should be an or

NybergWISC (commented):
Looks good to me overall. There are a few definitions that we should check with people who know more about fusion systems generally and comments/tooltips to clean up but great job getting version 0 up and running so quickly!

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+
+  /// The handleTick function specific to the Reactor.
+  /// @param time the time of the tock
+  virtual void Tock();
+
+  #pragma cyclus var { \
+    "doc": "Nameplate fusion power of the reactor", \
+    "tooltip": "Nameplate fusion power", \
+    "units": "MW", \
+    "uilabel": "Fusion Power" \
```
nuclearkatie, line 69: See comment below. Here are some example uitype implementations. Feel free to commit them if they're useful  or modify them if my suggestions aren't quite right. For example, assuming the power should be positive
```suggestion
    "uitype": "range", \
    "range": [0, 1e299], \
    "uilabel": "Fusion Power" \
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+    "tooltip": "Achievable system tritium breeding ratio before decay", \
+    "units": "non-dimensional", \
+    "uilabel": "Tritium Breeding Ratio" \
+  }
+  double TBR;
+
+  #pragma cyclus var { \
+    "doc": "Minimum tritium inventory to hold in reserve in case of tritium recovery system failure", \
+    "tooltip": "Minimum tritium inventory to hold in reserve (excluding core invneotry)", \
+    "units": "kg", \
```
nuclearkatie, line 84:
```suggestion
    "units": "kg", \
    "uitype": "range", \
    "range": [0, 1e299], \
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+    "uilabel": "Reserve Inventory" \
+  }
+  double reserve_inventory;
+
+  #pragma cyclus var { \
+    "doc": "Tritium core inventory required to start reactor", \
+    "tooltip": "Tritium core inventory required to start reactor", \
+    "units": "kg", \
```
nuclearkatie, line 92:
```suggestion
    "units": "kg", \
    "uitype": "range", \
    "range": [0, 1e299], \
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+    "uilabel": "Start-up Inventory" \
+  }
+  double startup_inventory;
+
+  #pragma cyclus var { \
+    "default": 'fill', \
+    "doc": "Method of refueling the reactor", \
+    "tooltip": "Options: 'schedule' or 'fill'", \
```
nuclearkatie, line 100:
```suggestion
    "tooltip": "Options: 'schedule' or 'fill'", \
    "uitype": "combobox", \
    "categorical": ['schedule', 'fill'], \
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  double buy_quantity;
+
+ #pragma cyclus var { \
+    "default": 1, \
+    "doc": "Frequency which reactor tries to purchase new fuel", \
+    "tooltip": "Reactor is active for 1 timestep, then dormant for buy_frequency-1 timesteps", \
+    "units": "Timesteps", \
+    "uilabel": "Buy frequency" \
+  }
+  double buy_frequency;
```
nuclearkatie, line 121: Is there a reason this is a double? Time steps are integers. What if your user submits a frequency of 1.5?

```suggestion
  int buy_frequency;
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  }
+  double min_blanket_enrichment;
+
+
+  //WARNING: The default on this is completely arbitrary!
+  #pragma cyclus var { \
+    "default": 0.05, \
+    "doc": "Percent of blanket that gets recycled every timestep", \
+    "tooltip": "Defaults to 5% (0.05), must be between 0 and 15%", \
+    "units": "%", \
```
nuclearkatie, line 174: Here's an example where using restrictions will help you enforce this limit you described
```suggestion
    "units": "%", \
    "uitype": "range", \
    "range": [0, 0.15], \
```

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  cyclus::toolkit::TotalInvTracker excess_tritium_tracker;
+
+  #pragma cyclus var {"tooltip":"Tracker to handle on-hand helium"}
+  cyclus::toolkit::TotalInvTracker helium_tracker;
+
+  cyclus::toolkit::MatlBuyPolicy fuel_startup_policy;
+  cyclus::toolkit::MatlBuyPolicy fuel_refill_policy;
+
+  //These two may be redundant (only one may be needed)
+  //cyclus::toolkit::MatlBuyPolicy blanket_startup_policy;
```
nuclearkatie, line 213: Determine if these are both needed and remove excess code (and excess comment)

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  cyclus::toolkit::MatlBuyPolicy blanket_fill_policy;
+
+  cyclus::toolkit::MatlSellPolicy tritium_sell_policy;
+  cyclus::toolkit::MatlSellPolicy helium_sell_policy;
+
+  //For the schedule buy policy
+  //cyclus::IntDistribution::Ptr active_dist = NULL;
+  //cyclus::IntDistribution::Ptr dormant_dist = NULL;
+  //cyclus::IntDistribution::Ptr size_dist = NULL;
```
nuclearkatie, line 222: Same as above. Either uncomment or remove

src/reactor.cc
nuclearkatie: And this should be in the header, you don't need to re-set this variable every time `DepleteBlanket` gets called

nuclearkatie (commented):
Noted a few administrative things like defaults and extra comments left in the code, and made a few suggestions about how to use uitypes to restrict the values your users can submit

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  /// @param time the time of the tock
+  virtual void Tock();
+
+  #pragma cyclus var { \
+    "doc": "Nameplate fusion power of the reactor", \
+    "tooltip": "Nameplate fusion power", \
+    "units": "MW", \
+    "uilabel": "Fusion Power" \
+  }
+  double fusion_power;
```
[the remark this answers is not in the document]
nuclearkatie: See the [preprocessor page](http://fuelcycle.org/arche/cycpp.html) for a bit more about the uitypes as annotation directives

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+  RecordOperationalInfo("Bred Tritium", std::to_string(bred_fuel->quantity()) +
+                                            " kg of T bred from blanket");
+
+  return bred_fuel;
+}
+
+void Reactor::OperateReactor(double TBR, double burn_rate) {
+  int seconds_per_year = 31536000;
+  double fuel_usage =
+      burn_rate * (fusion_power / 1000) / seconds_per_year * context()->dt();
```
gonuke, line 360: Magic number: what is 1000 here? Assign it to a variable (I assume its MW2kW)

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  cyclus::toolkit::MatlSellPolicy tritium_sell_policy;
+  cyclus::toolkit::MatlSellPolicy helium_sell_policy;
+
+  //For the schedule buy policy
+  //cyclus::IntDistribution::Ptr active_dist = NULL;
+  //cyclus::IntDistribution::Ptr dormant_dist = NULL;
+  //cyclus::IntDistribution::Ptr size_dist = NULL;
+
+  void Startup();
+  void OperateReactor(double TBR, double burn_rate=55.8);
```
gonuke, line 225: Magic number: where does 55.8 come from?

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+///
+/// @section optionalparams Optional Parameters
+/// Place a description of the optional input parameters to define the
+/// agent implementation.
+///
+/// @section detailed Detailed Behavior
+/// Place a description of the detailed behavior of the agent. Consider
+/// describing the behavior at the tick and tock as well as the behavior
+/// upon sending and receiving materials and messages.
+class Reactor : public cyclus::Facility  {
```
gonuke, line 34: You may want to call this a `FusionPowerPlant` just to be clear that it's different from a fission `Reactor`

gonuke (changes requested):
Here's a first pass.  Definitely should have been talking about the meat of your modeling choices, but maybe you've been doing that with others?

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+
+  //WARNING: The default on this is completely arbitrary!
+  #pragma cyclus var { \
+    "default": 0.05, \
+    "doc": "Percent of blanket that gets recycled every timestep", \
+    "tooltip": "Defaults to 5% (0.05), must be between 0 and 15%", \
+    "units": "%", \
+    "uilabel": "Blanket Turnover Rate" \
+  }
+  double blanket_turnover_rate;
```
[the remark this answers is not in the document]
NybergWISC: Just to reiterate after the conversation today, I think Version 1 will end up needing to change this behavior to reflect varied designs, schedules, downtimes, or requirements but something like this is fine for v0. 

The only part I would still like to talk about is the flexibility of timesteps we plan to have. As I understand, there would be reasons for simulations to change the timestep, so this variable (and other ones similarly defined) should scale based on a set timestep.

>

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void Reactor::DecayInventory(
+    cyclus::toolkit::ResBuf<cyclus::Material>& inventory) {
+  if (!inventory.empty()) {
+    cyclus::Material::Ptr mat = inventory.Pop();
+    mat->Decay(context()->time());
+    inventory.Push(mat);
+  }
+}
```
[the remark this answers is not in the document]
bennibbelink: Adding on this.... this code looks like it assumes the inventory has at most one material.  Also consider using `ResBuf::Peek()` instead of popping and pushing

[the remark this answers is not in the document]
gonuke: I think the way this archetype is using the `ResBuf` ensures that there is always only a single `Material` object in it.  That's part of my reason for suggesting a special version/behavior of `ResBuf`

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+  DepleteBlanket(fuel_usage * TBR);
+  cyclus::Material::Ptr mat = blanket.Pop();
+
+  cyclus::CompMap c = mat->comp()->mass();
+  cyclus::compmath::Normalize(&c, mat->quantity());
+
+  cyclus::CompMap T = {{10030000, 1}};
+
+  cyclus::Material::Ptr bred_fuel =
+      mat->ExtractComp(c[10030000], cyclus::Composition::CreateFromAtom(T));
```
gonuke, line 348: Make this composition once and store as a static member variable

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+
+  bool operational = true; 
+  bool core_loaded = false;
+
+  #pragma cyclus var {"tooltip":"Buffer for handling tritium kept in the core"}
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_core;
```
gonuke, line 183: It's not clear that this needs to be a `ResBuf` since its completely internal and always a single material

gonuke (changes requested):
A bunch of comments on code readability here.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+  RecordInventories(tritium_core.quantity(), tritium_reserve.quantity(),
+                    tritium_storage.quantity(), blanket.quantity(),
+                    helium_storage.quantity());
+}
+
+void Reactor::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_startup_policy
+      .Init(this, &tritium_reserve, std::string("Tritium Storage"),
```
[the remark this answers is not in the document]
nuclearkatie: Yeah, this results in behavior identical to initializing without an inventory policy. Material is always ordered whenever less than reserve+startup is on-hand, up to reserve+startup. This is equivalent to a max inventory size of reserve+startup without any throughput restrictions (the space remaining is always the request size)

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+  blanket_fill_policy
+      .Init(this, &blanket, std::string("Blanket Startup"), &blanket_tracker,
+            std::string("ss"), blanket_size, blanket_size)
+      .Set(blanket_incommod)
+      .Start();
+
+  // Tritium Buy Policy Selection:
+  if (refuel_mode == "schedule") {
+    cyclus::IntDistribution::Ptr active_dist =
+        cyclus::FixedIntDist::Ptr(new cyclus::FixedIntDist(1));
```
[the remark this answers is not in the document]
nuclearkatie: I think if you're specifying a dormant dist (which is true here), it probably makes sense to specify an active dist too since they're linked, even if it is equivalent to the default. To me, it makes more sense to pass `active_dist = NULL` and `dormant_dist = NULL` when you're only specifying a buying size dist

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+    "uilabel": "Fuel input commodity" \
+  }
+  std::string fuel_incommod;
+
+  #pragma cyclus var { \
+    "doc": "Fresh fuel commodity", \
+    "tooltip": "Name of fuel commodity requested", \
+    "uilabel": "Fuel input commodity" \
+  }
+  std::string he3_outcommod;
```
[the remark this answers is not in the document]
dean-krueger: This is mainly a product of my early understanding of how cyclus worked. Other agents have a more general input commodity "fuel" which can be things like UOX or MOX, and my thinking was that we wanted the same sort of thing. Currently the code only supports pure Tritium as fuel, and checks to make sure that this is what is supplied before running. However, I think leaving this as something the user can define is better looking towards the future, since even though it currently has to be T, eventually we'll want to be able to run off of a user defined mixture of D and T, but I'm definitely not married to that implementation.

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+    "uilabel": "Fuel input commodity" \
+  }
+  std::string blanket_incommod;
+
+  #pragma cyclus var { \
+    "doc": "Fresh fuel commodity", \
+    "tooltip": "Name of fuel commodity requested", \
+    "uilabel": "Fuel input commodity" \
+  }
+  std::string blanket_inrecipe;
```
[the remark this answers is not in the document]
dean-krueger: Same as above, currently it's restricted to being Li, but I made it general thinking towards the future when that will not be the case.

src/reactor.h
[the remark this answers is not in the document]
dean-krueger: I agree that this is potentially dangerous for weird timesteps, and that we should make the behavior more sophisticated for version 1. My thinking for leaving it in for this version is that ultimately it's user definable, and my understanding is that currently CYCLUS cannot change timestep size mid-simulation (though I believe this is a planned feature?), so if the behavior is weird the number can always be adjusted by the use to match the timestep.

Definitely worth keeping an eye on, I'll open an issue to address this once this PR goes through (not sure I can open the issue now, but I may be wrong).

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: Adding specific isotopes to the mixture would be fairly simple. As per our conversation in the ines-fusion meeting, I think adding a more sophisticated way of topping off the blanket is a good feature for version 1. I will make another issue note for when the PR goes through. Let me know if you disagree, and we can discuss further.

src/reactor.cc
dean-krueger: Good catch! Thanks

src/reactor.cc
dean-krueger: I went back and looked at the language that Abdou uses, and it seems like the way to think about it is that the _initial_ storage inventory is the startup, and includes the reserve (that is to say that you are correct). It seems like I misunderstood this when I first started implementation and never corrected it. I think I will need to go back and do a bit of an overhaul on my current three tritium inventories system and squash them into just one to make this more in line with Abdou. I will leave another comment when I've committed that change. This is a good catch, thanks!

src/reactor.cc
dean-krueger: Agreed, working on a fix to this now. Expect a commit addressing this soon.

src/reactor.cc
dean-krueger: Made Li7_contribution a cyclus var (user definable) and as per our discussion in ines-fusion, changed the default value from 0.15 to 0.03 to reflect the group consensus that 15% was too high.

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+/// agent implementation.
+///
+/// @section optionalparams Optional Parameters
+/// Place a description of the optional input parameters to define the
+/// agent implementation.
+///
+/// @section detailed Detailed Behavior
+/// Place a description of the detailed behavior of the agent. Consider
+/// describing the behavior at the tick and tock as well as the behavior
+/// upon sending and receiving materials and messages.
```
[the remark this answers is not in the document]
dean-krueger: Fixed. Thanks!

src/reactor.h
```diff
@@ -0,0 +1,242 @@
...
+  /// The Prime Directive
+  /// Generates code that handles all input file reading and restart operations
+  /// (e.g., reading from the database, instantiating a new object, etc.).
+  /// @warning The Prime Directive must have a space before it! (A fix will be
+  /// in 2.0 ^TM)
+
+  #pragma cyclus
+
+  #pragma cyclus note {"doc": "A stub facility is provided as a skeleton " \
+                              "for the design of new facility agents."}
```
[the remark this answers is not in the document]
dean-krueger: Fixed. Thanks!

src/reactor.h
dean-krueger: This is a good suggestion. I will go through and add these where they make sense. I just didn't think to do so initially.

src/reactor.h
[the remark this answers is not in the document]
dean-krueger: Added uitypes, good catch. Thanks!

src/reactor.h
dean-krueger: Fixed. Thanks!

src/reactor.h
dean-krueger: Added suggested UI type and changed language from "percent" to "fraction" as suggested.

src/reactor.h
dean-krueger: Did this one manually because I wasn't sure if it would break one of my unit tests, but this was implemented all the same. Thanks for spelling this one out, too, I couldn't quite figure out how to do it from just reading the documentation.

src/reactor.h
dean-krueger: Fixed. Thanks!

src/reactor.h
dean-krueger: Fixed. Thanks!

README.rst
```diff
@@ -0,0 +1,110 @@
+.. _hello_world:
+
+Hello, Cyclus!
```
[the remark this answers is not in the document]
dean-krueger: Changed from the default cyclus readme. The current readme is still a placeholder, but it should do for now.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+
+    RecordOperationalInfo(
+        "Tritium Moved",
+        std::to_string(core_deficit) + "kg of T moved from reserve to core");
+    if (surplus > 0.0) {
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from reserve to storage");
+    }
+    tritium_reserve.Push(reserve_fuel);
```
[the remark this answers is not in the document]
dean-krueger: Agreed. That would be a really nice feature.

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: I didn't know about ResBuf::Peek(), but yes, the archetype ensures taht there is always only a single material object in each resbuf (except for the brief moment between when the DRE adds one and the tock method squashes the ResBuf).

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+    cyclus::Material::Ptr base = inventory.Pop();
+    int count = inventory.count();
+    for (int i = 0; i < count; i++) {
+      cyclus::Material::Ptr m = inventory.Pop();
+      base->Absorb(m);
+    }
+
+    inventory.Push(base);
+  }
+}
```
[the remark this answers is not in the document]
dean-krueger: Agreed, that would be nice. Speaking of which, I'll take a look at the PRs you added in a minute.

gonuke:
Putting this before your ascii art will make it format correctly, I think:
```
.. code-block:: bash
```

README.rst
gonuke, line 1: I don't think you need both of these files (rst and md)

src/reactor.cc
```diff
@@ -0,0 +1,402 @@
...
+#include "boost/shared_ptr.hpp"
+
+namespace tricycle {
+
+Reactor::Reactor(cyclus::Context* ctx) : cyclus::Facility(ctx) {
+  // capacities set somewhat arbitrarily
+  fuel_tracker.Init({&tritium_reserve}, 1000.0);
+  blanket_tracker.Init({&blanket}, 100000.0);
+  excess_tritium_tracker.Init({&tritium_storage}, 1000.0);
+  helium_tracker.Init({&helium_storage}, 1000.0);
```
gonuke, line 12: magic numbers.  Should these be user inputs?

gonuke (changes requested):

src/reactor.cc
dean-krueger: This has been changed (and holy cow was it a bit bigger of a change than I thought it would be, dang!). I expect this will need another look (not necessarily by you, since you may be gone/busy with other stuff) in the next round of review.

src/reactor.cc
dean-krueger: I think because of the inclusion of Li7_contribution in the equation, the && works (on the edges at 0 and 1 it will drop the unneeded term because of a multiplication by either Li7_contribution or 1-Li7_contribution), but I definitely think adding a test to make sure that the values of bred tritium are consistent is a good idea.

dean-krueger: Done.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+    tritium_core.Push(core_fuel);
+
+    CombineInventory(tritium_storage);
+  }
+
+  // This pulls out some of the blanket each timestep so that fresh blanket can
+  // be added.
+  if (!blanket.empty() &&
+      blanket.quantity() >= blanket_size * blanket_turnover_rate) {
+    cyclus::Material::Ptr blanket_mat = blanket.Pop();
```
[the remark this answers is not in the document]
dean-krueger: It looks like you can! I didn't notice that when I was writing this initially. Thanks!

src/reactor.h
dean-krueger: I added a comment with a citation and explanation about what this value is and where it comes from.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+  double Li7_contribution = 0.15;
+
+  cyclus::CompMap depleted_comp;
+
+  // This is ALMOST the correct behavior, but "scraping the bottom of the
+  // barrel" is a little too complex for this implementation.
+  if ((b[30060000] - (1 - Li7_contribution) * 2 * bred_tritium_mass > 0) &&
+      (b[30070000] - Li7_contribution * 7.0 / 3.0 * bred_tritium_mass > 0)) {
+    depleted_comp = {{30070000, b[30070000] - Li7_contribution * 7.0 / 3.0 *
+                                                  bred_tritium_mass},
```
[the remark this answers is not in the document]
dean-krueger: I could be wrong, but because this references b[30070000] and b[30060000] (the quantity of Li6 and Li7 in the blanket), and the blanket composition changes over time as it gets depleted and recycled, it is necessary to recalculate this value every time more tritium is bred. This is also why there's an if statement before it checking whether there's enough Li6 and Li7 left in the blanket to proceed with the depletion.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+    RecordOperationalInfo(
+        "Tritium Moved",
+        std::to_string(core_deficit) + "kg of T moved from reserve to core");
+    if (surplus > 0.0) {
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from reserve to storage");
+    }
+    tritium_reserve.Push(reserve_fuel);
+    tritium_core.Push(core_fuel);
```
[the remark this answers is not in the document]
dean-krueger: Changed how this worked overall as a part of a different modification, if it's still weird on a second review (with the updated code) I'll gladly change it further, though.

src/reactor.cc
dean-krueger: First two points addressed, last point will be addressed shortly when I make the big kg --> atoms change and make everything more precise.

src/reactor.h
dean-krueger: Addressed. This got deleted in favor of the new way of doing things as per Matthew's comment about making the code match Abdou 2021 better.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+    blanket.Push(blanket_mat);
+  } else if (!blanket.empty() &&
+             blanket.quantity() < blanket_size * blanket_turnover_rate) {
+    RecordOperationalInfo(
+        "Blanket Not Cycled",
+        "Total blanket material (" + std::to_string(blanket.quantity()) +
+            ") insufficient to extract " +
+            std::to_string(blanket_size * blanket_turnover_rate) + "kg!");
+  }
+}
```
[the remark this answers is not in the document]
dean-krueger: Good call, that's a lot better. Done.

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: Is this something that needs to be changed, or just a comment that it was a funky way for me to have implemented it? I ask because this was the way that seemed the most obvious to me, and my understanding is that it works the way I want it to (as long as inventory < startup+reserve, try to buy however much fuel will get inventory = startup+reserve). It's worth noting that I've changed the specific requirements on this a little bit, but the overall way it works is still the same. Happy to change it if it needs to go, but I'm not totally sure what I would change it to if I need to, so just checking.

README.rst
dean-krueger: Deleted .md since github seems to prefer .rst files. I agree, two was unnecessary.

gonuke: Actually - I think GitHub prefers MD

src/reactor.cc
dean-krueger: Decided to not convert burn_rate to atoms, just because I found a different way of doing it which made more sense to me, but which I think achieves the same result. This may have been "the hard way", who knows?

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+
+void Reactor::ExtractHelium(
+    cyclus::toolkit::ResBuf<cyclus::Material>& inventory) {
+  if (!inventory.empty()) {
+    cyclus::Material::Ptr mat = inventory.Pop();
+    cyclus::CompMap c = mat->comp()->atom();
+    cyclus::compmath::Normalize(&c, mat->quantity());
+
+    cyclus::CompMap He3 = {{20030000, 1}};
```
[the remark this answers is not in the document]
dean-krueger: Done, as per our discussion today in software meeting.

src/reactor.cc
dean-krueger: Done.

src/reactor.cc
```diff
@@ -0,0 +1,406 @@
...
+
+  // Percent (as decimal) of T which comes from Li-7 instead of Li-6
+  double Li7_contribution = 0.15;
+
+  cyclus::CompMap depleted_comp;
+
+  // This is ALMOST the correct behavior, but "scraping the bottom of the
+  // barrel" is a little too complex for this implementation.
+  if ((b[30060000] - (1 - Li7_contribution) * 2 * bred_tritium_mass > 0) &&
+      (b[30070000] - Li7_contribution * 7.0 / 3.0 * bred_tritium_mass > 0)) {
```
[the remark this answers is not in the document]
dean-krueger: Done, good suggestion. I had this in place from my first pass at the calc where I was just trying to get it approximately right without thinking about it much. Definitely needed to change. It should be fixed now. Thanks!

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: Fixed.

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
+       it != c.end(); ++it) {
+    comp = comp + std::string("{") + std::to_string(it->first) +
+           std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
+}
```
gonuke, line 158: Seems like motivation for a `Composition::str()` method

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+    tritium_sequestered.Push(tritium_storage.Pop(sequestered_equilibrium));
+  }
+}
+
+void Reactor::Startup() {
+  cyclus::Material::Ptr initial_storage = tritium_storage.Peek();
+  cyclus::CompMap c = initial_storage->comp()->atom();
+  cyclus::compmath::Normalize(&c, 1);
+
+  if ((tritium_storage.quantity() >= (startup_inventory)) &&
```
gonuke, line 185:
```suggestion
  if ((tritium_storage.quantity() >= startup_inventory) &&
```

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+      (GetComp(initial_storage) == expected_fuel_comp) &&
```
gonuke, line 186: Why a string comparison? It is preferable to just compare compositions here as `Composition` objects.

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+void Reactor::SequesterTritium(){
+  if (!tritium_sequestered.empty()){
+    cyclus::Material::Ptr sequestered_mat = tritium_sequestered.Pop();
+    cyclus::CompMap c = sequestered_mat->comp()->atom();
+    cyclus::compmath::Normalize(&c, sequestered_mat->quantity());
+
+    double equilibrium_deficit = sequestered_equilibrium - c[tritium_id];
+
+    // Another catch might be good here...
+    sequestered_mat->Absorb(tritium_storage.Pop(equilibrium_deficit));
```
gonuke, line 169: Could `equilibrium_deficit` end up negative here? Seems like a problem...

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+
+    cyclus::Material::Ptr storage_fuel = tritium_storage.Pop();
+    tritium_excess.Push(storage_fuel->ExtractQty(surplus));
+
+    if (surplus > 0.0) {
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from storage to excess");
+    }
+    tritium_storage.Push(storage_fuel);
```
gonuke, line 47: You should be able to `Pop(surplus)` and avoid having to extract and push back onto the buffer.

Also only necessary if `surplus > 0`

```suggestion

    if (surplus > 0.0) {

      tritium_excess.Push(tritium_storage.Pop(surplus));

      RecordOperationalInfo(
          "Tritium Moved",
          std::to_string(surplus) + "kg of T moved from storage to excess");
    }
```

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+  }
+
+  // This pulls out some of the blanket each timestep so that fresh blanket can
+  // be added.
+  // NOTE: currently spent blanket dissapears into the ether. This needs to be
+  // changed in version 1 for materials tracking.
+
+  double blanket_turnover = blanket_size * blanket_turnover_rate;
+  if (!blanket.empty() && blanket.quantity() >= blanket_turnover && 
+      context()->time() % blanket_turnover_frequency == 0) {
```
gonuke, line 58: None of this should happen unless this is true first

src/reactor.cc
gonuke, line 186: String comparison should be replaced with `comp math::AlmostEq()`

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+  } else {
+    // Should I do a try/catch here, or is "user error" fine?
+    tritium_sequestered.Push(tritium_storage.Pop(sequestered_equilibrium));
+  }
+}
+
+void Reactor::Startup() {
+  cyclus::Material::Ptr initial_storage = tritium_storage.Peek();
+  cyclus::CompMap c = initial_storage->comp()->atom();
+  cyclus::compmath::Normalize(&c, 1);
```
gonuke, line 183: You never use `c`; is this all about normalizing `initial_storage`? Doesn't seem like that will work

src/reactor.cc
gonuke, line 185: Since all other branches result in throwing exceptions, do those first one at a time:
* not enough fuel
* less than usage
* wrong comp

then if you get through those, just do startup stuff

src/reactor.h
```diff
@@ -0,0 +1,342 @@
...
+
+  #pragma cyclus var {"tooltip":"Buffer for handling tritium material to be used in reactor"}
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_storage;
+
+  #pragma cyclus var {"tooltip":"Buffer for handling excess tritium material to be sold"}
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_excess;
+
+  #pragma cyclus var {"tooltip":"Buffer for handling tritium sequestered in the system"}
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_sequestered;
+
```
gonuke, line 267: These can just be `Material` - will make your life easier, I think.

src/reactor.h
```diff
@@ -0,0 +1,342 @@
...
+  void RecordInventories(double storage, double excess, double sequestered, double blanket, double helium);
+  void RecordOperationalInfo(std::string name, std::string val);
+  void DepleteBlanket(double bred_tritium_mass);
+  cyclus::Material::Ptr BreedTritium(double fuel_usage, double TBR);
+  std::string GetComp(cyclus::Material::Ptr mat);
+
+
+  private: 
+    bool sufficient_tritium_for_operation = false;
+    const double seconds_per_year = 31536000.0;
```
gonuke, line 305: Isn't this defined at the cyclus level?

src/reactor.h
```diff
@@ -0,0 +1,342 @@
...
+    const double MW_to_GW = 1000.0;
+    double fuel_usage;
+    std::string expected_fuel_comp = "{{10030000,1.000000}}";
+
+    //masses in kg/atom
+    const double Li7_atomic_mass = 1.16503478e-26;
+    const double Li6_atomic_mass = 9.9883465e-27;
+    const double tritium_atomic_mass = 5.00826766e-27;
+    const double He4_atomic_mass = 6.64647907e-27;
```
gonuke, line 314: See `pyne::atomic_mass()`

src/reactor.cc
gonuke: But you only use this for comparing compositions & there is a method for that: `compmath::AlmostEq()`

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
+}
+
+void Reactor::SequesterTritium(){
+  if (!tritium_sequestered.empty()){
+    cyclus::Material::Ptr sequestered_mat = tritium_sequestered.Pop();
+    cyclus::CompMap c = sequestered_mat->comp()->atom();
+    cyclus::compmath::Normalize(&c, sequestered_mat->quantity());
```
gonuke, line 164: See `MatQuery` in the toolkit for what you want here & in many other places

gonuke (changes requested):
A number of additional tips... you're reinventing the wheel in a few places

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+      ->AddVal("TritiumSequestered", sequestered)
+      ->AddVal("LithiumBlanket", blanket)
+      ->AddVal("HeliumStorage", helium)
+      ->Record();
+}
+
+void Reactor::DepleteBlanket(double bred_tritium_mass) {
+  cyclus::Material::Ptr blanket_mat = blanket.Pop();
+
+  cyclus::CompMap b = blanket_mat->comp()->mass();
```
gonuke, line 294: If you get the `atom()` version, doesn't it make your life easier?

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+
+
+  RecordInventories(tritium_storage.quantity(), tritium_excess.quantity(), tritium_sequestered.quantity(), blanket.quantity(),
+                    helium_storage.quantity());
+}
+
+void Reactor::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_usage = burn_rate * (fusion_power / MW_to_GW) / seconds_per_year * context()->dt();
```
gonuke, line 97: How does your code change if this is in atoms/moles and not kg?!?!  My hunch is that it gets simpler throughout.

gonuke (changes requested):
More thoughts

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+
+Reactor::Reactor(cyclus::Context* ctx) : cyclus::Facility(ctx) {
+  // capacities set somewhat arbitrarily
+  fuel_tracker.Init({&tritium_storage}, 1000.0);
+  blanket_tracker.Init({&blanket}, 100000.0);
+}
+
+std::string Reactor::str() {
+  return Facility::str();
+}
```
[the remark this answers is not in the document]
dean-krueger: This is auto generated when you make a new cyclus agent, and since I didn't exactly know what purpose it served, I decided it was better not to touch it... If it's something I should delete I'm happy to do it though!

src/reactor.cc
dean-krueger: I most certainly can! I did this manually since I also incorporated the suggestion below. Good catch!

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+    cyclus::Material::Ptr storage_fuel = tritium_storage.Pop();
+    tritium_excess.Push(storage_fuel->ExtractQty(surplus));
+
+    if (surplus > 0.0) {
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from storage to excess");
+    }
+    tritium_storage.Push(storage_fuel);
+    CombineInventory(tritium_excess);
```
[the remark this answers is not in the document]
dean-krueger: Changed. Good catch.

src/reactor.cc
```diff
@@ -0,0 +1,375 @@
...
+  }
+
+  // This pulls out some of the blanket each timestep so that fresh blanket can
+  // be added.
+  // NOTE: currently spent blanket dissapears into the ether. This needs to be
+  // changed in version 1 for materials tracking.
```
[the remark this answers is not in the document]
dean-krueger: Yeah, I'm honestly not sure why I didn't do that earlier. It's fixed now though.

src/reactor.cc
dean-krueger: Agreed that such a method would be nice. I use it in one of my error messages to show the user what the incorrect composition for fuel is, so I didn't remove the method entirely, but I did switch to the compmath::AlmostEq() method for the comparisons. It felt less janky that way. I didn't know that method existed.

src/reactor.cc
dean-krueger: Done, good point. I made the threshold 1e-7 because I didn't want to run into any weirdness with floating point math.

src/reactor.cc
dean-krueger: I couldn't think of a way for that to happen, but I am also not the final say in whether or not it's possible, so I added a catch (std::max(eq_def, 0.0). Definitely worth the extra typing to prevent bugs.

src/reactor.cc
dean-krueger: I'm not entirely sure what this is referring to, I'll be honest...

src/reactor.cc
dean-krueger: That is true, C wasn't getting used. Now it is, though, in AlmostEq().

src/reactor.cc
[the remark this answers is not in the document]
gonuke: It is autogenerated as a minimal stub.  If you want more than the minimum information printed about your agent when it is converted to a string, then you need to add it here.  

The first line `Facility::str()` ensures that this minimal amount is written. If you don't add anything else, I am 99% sure this is the same as removing it, as `Facility::str()` will be used if you don't override it yourself.

src/reactor.cc
gonuke: The last part of your conditional tests whether it is time to do blanket turnover.  I think this should be tested first by itself, and then the rest of the tests should only happen if it's true.

src/reactor.cc
dean-krueger: Yep, that's absolutely the right way to do that. Good point. Fixed

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: Ahh, gotcha. I'll try taking it out and see if it breaks anything. Worse case it shouldn't be too hard to add back down the road.

src/reactor.cc
dean-krueger: ohh, gotcha. After looking at it I agree. Change made. Good call. Thanks!

src/reactor.h
dean-krueger: Yep, didn't know that was a thing. Good suggestion!

src/reactor.cc
```diff
@@ -0,0 +1,373 @@
...
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from storage to excess");
+    }
+  }
+
+  // This pulls out some of the blanket each timestep so that fresh blanket can
+  // be added.
+  double blanket_turnover = blanket_size * blanket_turnover_rate;
+  if (context()->time() % blanket_turnover_frequency == 0) {
+    if (!blanket.empty() && blanket.quantity() >= blanket_turnover) {
```
gonuke, line 53: Since both conditions require `!blanket.empty()` maybe combine that with the first condition to simplify further?

gonuke (changes requested):
one simplifying suggestion for a conditional

src/reactor.h
dean-krueger: The number of seconds in a month is defined as the number of seconds in a year/12 I think, so I could use that*12 instead, but I checked the value and it's not quite 1/12 of a year (although how long a year is depends on who you ask really, NASA says 365.2422, some people say 365.25, some people say 365, none of those lead to a value which matches what's in cyclus which is 365.25639 days, or what appears to be a gaussian year, not totally sure since it doesn't match exactly, but maybe that's a "we wanted it to be an int" rounding thing). I do not think cyclus defines the seconds in a year explicitly, at least not in context. Not sure the paragraph was totally necessary, I'll change this to context()->kDefaultTimeStepDur*12 for consistency.

dean-krueger: I don't think I can access that variable, actually, since it never actually makes its way into context, and is just defined in the header and used as a default. I can't just use dt() since it's possible to change from the default, too. Just going to make it the same number times 12 for consistency. If I'm missing something, though, feel free to let me know.

src/reactor.cc
dean-krueger: Yep. You're totally right. Done.

src/reactor.cc
dean-krueger: I can't make them user inputs (the inventory trackers have to be initialized in the constructor, which it looks like goes before the user inputs are processed). I just made them member variables. They're meant to be ceilings on values anyway, so they're just set to sort of unreasonably high numbers.

src/reactor.h
dean-krueger: Changed sequestered_tritium to a material. tritium_excess needs to be a buffer since it's how we sell off excess tritium (there's a policy attached). Checked all the other buffers to see if they could be converted, and the only other one was blanket_excess, which should eventually have a policy added (in a future version), so I'm going to leave that one as a buffer for now.

src/reactor.cc
```diff
@@ -0,0 +1,373 @@
+#include "reactor.h"
+
+#include "boost/shared_ptr.hpp"
+
+namespace tricycle {
+
+Reactor::Reactor(cyclus::Context* ctx) : cyclus::Facility(ctx) {
+  // capacities set somewhat arbitrarily
+  fuel_tracker.Init({&tritium_storage}, 1000.0);
+  blanket_tracker.Init({&blanket}, 100000.0);
```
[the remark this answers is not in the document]
dean-krueger: Fixed. Couldn't make user input, but they're no longer just floating around in there.

src/reactor.cc
dean-krueger: Fixed. Had to leave things the old way (compmap c --> normalize(c, 1)) in startup because matquary doesn't quite do what I need it to there, but everywhere else it should be gone (I'll be removing GetComp() in a minute)

src/reactor.cc
dean-krueger: Fixed. This ultimately revealed a bug, which I fixed as well.

src/reactor.cc
dean-krueger: Yes. Fixed.

src/reactor.cc
```diff
@@ -0,0 +1,332 @@
...
+  cyclus::toolkit::MatQuery b(blanket_mat);
+  cyclus::CompMap depleted_comp;
+
+  double converted_Li6 = Li6_contribution * bred_tritium_atoms;
+  double converted_Li7 = Li7_contribution * bred_tritium_atoms;
+  double bred_He4 = bred_tritium_atoms;
+  double blanket_Li6 = b.moles(Li6_id)*avagadros_number;
+  double blanket_Li7 = b.moles(Li7_id)*avagadros_number;
+  double blanket_tritium = b.moles(tritium_id)*avagadros_number;
+  double blanket_He4 = b.moles(He4_id)*avagadros_number;
```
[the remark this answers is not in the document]
gonuke: and now I wonder if you can do this on a mole basis instead of atoms....

gonuke (changes requested):
Definitely looking cleaner and more readable! A few more quick comments while you are in the middle of edits.

src/reactor.cc
```diff
@@ -0,0 +1,332 @@
...
+
+  double blanket_turnover = blanket_size * blanket_turnover_rate;
+  if (context()->time() % blanket_turnover_frequency == 0 && !blanket.empty()) {
+    if (blanket.quantity() >= blanket_turnover) {
+      blanket_excess.Push(blanket.Pop(blanket_turnover));
+      CombineInventory(blanket_excess);
+      RecordOperationalInfo(
+          "Blanket Cycled",
+          std::to_string(blanket_turnover) + "kg of blanket removed");
+    } else if (blanket.quantity() < blanket_turnover) {
```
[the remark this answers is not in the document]
dean-krueger: It sure does. Good catch, thanks!

src/reactor.cc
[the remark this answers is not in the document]
dean-krueger: Combined those variables, and yes. Yes you can do this on a mole basis. It cleans up some of the multiplying by Avagadro's number (though ultimately you do still have to multiply/divide by it once or twice since at the end of the day you need mass).

src/reactor.cc
```diff
@@ -0,0 +1,332 @@
...
+      tritium_excess.Push(tritium_storage.Pop(surplus));
+      CombineInventory(tritium_excess);
+
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from storage to excess");
+    }
+  }
+
+  double blanket_turnover = blanket_size * blanket_turnover_rate;
```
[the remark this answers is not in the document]
dean-krueger: yep! Done.

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
...
+  RecordInventories(tritium_storage.quantity(), tritium_excess.quantity(), 
+                    sequestered_tritium->quantity(), blanket.quantity(),
+                    blanket_excess.quantity(), helium_storage.quantity());
+}
+
+void Reactor::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / seconds_per_year * context()->dt());
+  fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
```
gonuke, line 85: since you've switched to moles, why not make this moles instead of atoms, and all of your atomic masses can be molar masses....

gonuke (changes requested):
continued march towards a simpler chemistry math - all-in on moles

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
+#include "reactor.h"
+
+#include "boost/shared_ptr.hpp"
+
```
gonuke, line 4: consider including some of these to reduce the need to fully specify the namespace of frequently used cyclus classes:
```
using cyclus::Material;
using cyclus::Composition;
using cyclus::toolkit::ResBuf;
using cyclus::toolkit::MatVec;
using cyclus::KeyError;
using cyclus::ValueError;
using cyclus::Request;
using cyclus::CompMap;
```

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
...
+}
+
+void Reactor::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / seconds_per_year * context()->dt());
+  fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
+  blanket_turnover = blanket_size * blanket_turnover_rate;
+
+  fuel_startup_policy
```
gonuke, line 88: We should discuss all these policies at a fuel cycle meeting with @nuclearkatie present.  My hunch is that there is an easier way to accomplish what you want.

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
...
+    if (surplus > 0.0) {
+
+      tritium_excess.Push(tritium_storage.Pop(surplus));
+      CombineInventory(tritium_excess);
+
+      RecordOperationalInfo(
+          "Tritium Moved",
+          std::to_string(surplus) + "kg of T moved from storage to excess");
+    }
+  }
```
gonuke, line 42: Maybe put this in a function called `StoreExcessTritium()` for readability.

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
...
+    cyclus::compmath::Normalize(&depleted_comp, 1);
+
+    // Because there's a mass difference between T+He and Li we need a new mass
+    double new_mass = ((remaining_Li7) * Li7_atomic_mass
+                    + (remaining_Li6) * Li6_atomic_mass
+                    + (blanket_tritium + bred_tritium_moles) * tritium_atomic_mass
+                    + (blanket_He4 + bred_He4) * He4_atomic_mass)*avagadros_number;
+
+    blanket_mat = cyclus::Material::Create(this, new_mass,
+                    cyclus::Composition::CreateFromAtom(depleted_comp));
```
gonuke, line 280: This is cyclus-dangerous since your old material just vaporizes and you create a new one from nothing here with a different id

src/reactor.cc
```diff
@@ -0,0 +1,333 @@
...
+        "Breeding Error",
+        "Blanket composition lacks sufficient lithium to continue "
+        "breeding at perscribed rate");
+  }
+  blanket.Push(blanket_mat);
+}
+
+cyclus::Material::Ptr Reactor::BreedTritium(double atoms_burned, double TBR) {
+  DepleteBlanket(atoms_burned/avagadros_number * TBR);
+  cyclus::Material::Ptr mat = blanket.Pop();
```
gonuke, line 295: The last thing that `DepleteBlanket()` does is push a material onto the buffer and then the first thing you do here is pop it off the buffer again.  Maybe `DepleteBlanket()` should just return the material ptr?

gonuke (changes requested):
Now that the code is more readable, we should revisit your modeling choices and function names.

src/reactor.cc
gonuke: Similarly, the next block could be `StoreBlanketTritium()` or something like that??

src/reactor.cc
dean-krueger: I had a bad feeling that this was illegal, but it seemed like the most straightforward way to get this to work. I'll come up with a different way to do this.

gonuke: We should look at what the RecipeReator does in Cycamore. I may be overly cautious here, but it seems like it will be hard to do any material accountability if materials just cease to exist

src/reactor.cc
```diff
@@ -0,0 +1,343 @@
...
+  }
+}
+
+void Reactor::Startup() {
+  cyclus::Material::Ptr initial_storage = tritium_storage.Peek();
+  cyclus::CompMap c = initial_storage->comp()->atom();
+  cyclus::compmath::Normalize(&c, 1);
+
+  if (tritium_storage.quantity() < startup_inventory){
+    throw cyclus::ValueError(
```
nuclearkatie, line 152: Do you actually want this to throw a ValueError and break the simulation, or just not start up at this point?

Issue #3, opened 2024-08-06
Define the resource buffers needed by the fusion power plant


gonuke:
Add the definition of the resource buffers that will be used in the fusion power plant archetype, and associated initialization variables that should be included in the archetype's definition via the `#pragma cyclus` calls.

dean-krueger:
I guess I used the wrong keyword in PR #9. Closing this issue as complete.

gonuke:
I think it likes "closes" or "fixes" as verbs

gonuke:
Revisiting this after some discussion:

Feed inventories - material can be purchased from outside:
* blanket feed
* tritium feed

Product inventories - material can be sold to outside:
* tritium product
* he-3 product

Waste inventories - material can be transferred to outside:
* blanket waste

Active inventories - not engaged in outside trading - could be just material objects
* core blanket
* core tritium
* sequestered tritium

Am I missing any, @dean-krueger ?

dean-krueger:
I think that about sums it up for what we had in mind for this version of FPP, with some other options for later versions being things like Deuterium feed/product, etc (which we don't need to implement right now I think). The one thing I'll say is that I think it makes sense for the concept of the "Core" to be transient. For example, things might live inside their appropriate buffers and then only get "pulled into the core" for operations that need to be performed on them, such as breeding, burning, transmuting, etc, before they're moved back into to the appropriate place. That may be a larger discussion, though?

gonuke:
I'm not sure how transient you mean... if your blanket is only turning over occasionally, then material should reside there. Each turnover should remove some to waste and replace it with feed. I don't think blanket material should ever flow from core back to feed.

PR #9, opened 2024-08-26
Added resource buffers

dean-krueger:
Added the resource buffers from the original tricycle implementation, since I think we'll need all of them (unless we want to make big changes to how tricycle works, which we may). 

Added in some of the beginning comments in the .h file which describe the fpp archetype. Not all of the fields could be completed, though, since this is still really early on.

Added some necessary files (install.py, CMakeLists.txt, src/CMakeLists.txt) for fpp to build.

Added some print statements to fpp.cc for testing, and feel like they're worth leaving in until we put something real there since otherwise I'm going to have to add them back and then delete them every time I want to see if things are working.

I took out the #pragma cyclus var {"capacity" : "1000"} lines from the resbuf definitions because they were arbitrary limits and I found out you a. don't need them for it to work, because b. there's a built in default of 1e300 which should be fine. 

One thing to note: If I remember correctly, there was some issue with the original implementation where if the user had more than some amount of tritium in their resource buffer the bit of code that extracts the helium broke, but since we're reworking this from the ground up that may be solved again later. Just wanted to mention it so that we keep our eyes open for that potential bug later on.

This PR addresses #3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,138 @@
+# DO NOT DELETE THIS LINE:: .:tricycle:fpp
+
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
+# taken from http://geant4.cern.ch/support/source/geant4/CMakeLists.txt
+IF(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
+    message(STATUS "Stub requires an out-of-source build.")
+    message(STATUS "Please remove these files from ${CMAKE_BINARY_DIR} first:")
+    message(STATUS "CMakeCache.txt")
+    message(STATUS "CMakeFiles")
+    message(STATUS "Once these files are removed, create a separate directory")
+    message(STATUS "and run CMake from there")
+    message(FATAL_ERROR "in-source build detected")
+ENDIF()
+
+PROJECT(tricycle)
+
+# check for and enable c++11 support
+INCLUDE(CheckCXXCompilerFlag)
+CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+IF(COMPILER_SUPPORTS_CXX11)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+ELSE()
+  MESSAGE(STATUS "The compiler ${CMAKE_CXX_COMPILER} has no C++11
+  support. Please use a different C++ compiler.")
+ENDIF()
+
+# quiets fortify_source warnings when not compiling with optimizations
+# in linux distros where compilers were compiled with fortify_source enabled by
+# default (e.g. Arch linux).
+MESSAGE("${CMAKE_BUILD_TYPE}")
+STRING(TOLOWER "${CMAKE_BUILD_TYPE}" BUILD_TYPE)
+IF(NOT ${BUILD_TYPE} STREQUAL "release")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0")
+ENDIF()
+# no overflow warnings because of silly coin-ness
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-overflow")
+
+# Direct any out-of-source builds to this directory
+SET(STUB_SOURCE_DIR ${CMAKE_SOURCE_DIR})
+
+# Direct any binary installation paths to this directory
+SET(STUB_BINARY_DIR ${CMAKE_BINARY_DIR})
+
+# This makes all the libraries build as SHARED
+SET(BUILD_SHARED_LIBS true)
+
+# Setup build locations.
+IF(NOT CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+    SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${STUB_BINARY_DIR}/bin)
+endif()
+IF(NOT CMAKE_LIBRARY_OUTPUT_DIRECTORY)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
+# DO NOT DELETE THIS LINE:: .:tricycle:fpp
+
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
```
gonuke, line 3: I recommend we update this to a newer version - this may have been changed in Cyclus since you started?

CMakeLists.txt
gonuke, line 1: Do you get these automatically with cycstub?

gonuke: I wonder if that was updated with recent updates to Cyclus?

src/fpp.h
```diff
@@ -9,27 +9,28 @@ namespace tricycle {
 
 /// @class fpp
 ///
-/// This Facility is intended
-/// as a skeleton to guide the implementation of new Facility
-/// agents.
-/// The fpp class inherits from the Facility class and is
+/// The Reactor class inherits from the Facility class and is
```
gonuke, line 12: But this is the FPP class, so shouldn't it still refer to `fpp`?

Now that I've approved the previous PR, I'm wondering if we actually want to use the full name "FusionPowerPlant"?

src/fpp.h
```diff
@@ -9,27 +9,28 @@ namespace tricycle {
...
 /// dynamically loaded by the Agent class when requested.
 ///
 /// @section intro Introduction
-/// Place an introduction to the agent here.
+/// This agnet is designed to function as a basic representation of a fusion
```
gonuke, line 16:
```suggestion
/// This agent is designed to function as a basic representation of a fusion
```

src/fpp.h
```diff
@@ -58,6 +59,13 @@ class fpp : public cyclus::Facility  {
   /// @param time the time of the tock
   virtual void Tock();
 
+  //Resource Buffers:
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_core;
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_reserve;
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_excess;
+  cyclus::toolkit::ResBuf<cyclus::Material> helium_storage;
+  cyclus::toolkit::ResBuf<cyclus::Material> blanket;
```
gonuke, line 67: Didn't you decide that some of these didn't need to be `ResBuf`?  That said, we have new `ResBuf` capabilities, so it may be OK to keep them.

gonuke (changes requested):
A few typos, but also some questions about how cycstub works with updated cyclus?

CMakeLists.txt
dean-krueger: They are all automatically generated with Cycstub, yes. I think I may have an old version of Cyclus on this machine (I wasn't able to ssh into the docker machine when I did this because it was turned off), but maybe if I start over again and do it through docker everything will be updated? I think now is a good time to redo everything, since we're so early on.

src/fpp.h
dean-krueger: This was my fault for just copying things over from the descriptions on the old version of Tricycle, and yeah, I think FusionPowerPlant is maybe better (more descriptive). This can be fixed pretty easy by just redoing cycstub and then adding the resbufs back in. Good catch.

src/fpp.h
dean-krueger: Now that you mention it, I remember having this conversation a while back. When I go through and redo cycstub today I'll also take a look at making some of these just floating materials instead of ResBufs.

CMakeLists.txt
dean-krueger: I thought that remaking everything with the newer version of Cyclus that exists on the CNERG-04 would do the trick, but it seems like it didn't. Not sure how to do this otherwise. Can I just change that line to have a newer version? I'd try it out, but worry about potential unforeseen, subtle downstream effects if I do without bringing it up.

gonuke: That may mean that we didn't update cycstub generation properly.  How do you run cycstub?

src/fusion_power_plant.h
```diff
@@ -46,21 +47,30 @@ class fpp : public cyclus::Facility  {
...
   virtual void Tock();
 
+  //Member Variables:
+
+  //Resource Buffers:
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_storage;
+  cyclus::toolkit::ResBuf<cyclus::Material> tritium_excess;
+  cyclus::toolkit::ResBuf<cyclus::Material> helium_excess;
+  cyclus::toolkit::ResBuf<cyclus::Material> blanket_storage;
+  cyclus::toolkit::ResBuf<cyclus::Material> blanket_excess;
```
gonuke, line 69: Reposting the question about ResBuf vs Material for this updated file

gonuke (changes requested):
Reposted the question about resbufs but need to look into cycstub

CMakeLists.txt
dean-krueger: Within the docker environment (with the newest version of cyclus, unless there was a release over the summer that I missed) I just do `cycstub .:tricycle:FusionPowerPlant`

CMakeLists.txt
gonuke: I have a branch to update Cycstub: cyclus/cyclus#1789 but the changes aren't that extensive

src/fusion_power_plant.h
dean-krueger: Resolved as of our conversation yesterday

CMakeLists.txt
dean-krueger: Made the changes from Cyclus PR #1789 and tested them out. We should be good on this now. Thanks for getting that going so quickly!

gonuke (approved):
This looks good... I haven't tested whether it builds - maybe we should add CI quickly.

PR #10, opened 2024-08-29
Pseudocode implementation of key functions

### LINKED ISSUE #4: Define major processes using pseudo-code in appropriate cyclus methods
In each of the main cyclus methods (`EnterNotify()`, `Tick()`, `Tock()`) use pseudo code to define the major actions that must take place to represent the fuel cycle of a functioning fusion power plant.  Confine this to methods that are about internal operations and not about trading resources.

dean-krueger:
Note: I made a bunch of comments and left things sort of rough since I figured part of the point of this exercise was to get feedback from you about the best way to do things. This represents the way it makes the most sense in my head, but I am obviously open to other ideas.

Adds Pseudocode implementations for Tick, Tock, and EnterNotify functions without supporting code in the .h file.
Removed print statements since this code won't run anyway as it is.

This PR Addresses #4 
This PR Closes #4

### DIFF
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -18,6 +18,10 @@ void FusionPowerPlant::EnterNotify() {
   fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
   blanket_turnover = blanket_size * blanket_turnover_rate;
 
+  //Create the blanket material for use in the core
+  const cyclus::Composition::Ptr enriched_li = context->GetRecipe(blanket_inrecipe);
+  cyclus::Material::Ptr blanket = cyclus::Composition::CreateFromAtom(0.0, enriched_li);
+
   //Truncated for readability. It goes on Tritium Storage and gets started, though.
   fuel_startup_policy.Init(&tritium_storage).Set(fuel_incommod).Start();
   
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -2,15 +2,15 @@
 
 namespace tricycle {
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 FusionPowerPlant::FusionPowerPlant(cyclus::Context* ctx) : cyclus::Facility(ctx) {}
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 std::string FusionPowerPlant::str() {
   return Facility::str();
 }
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::EnterNotify() {
   //Pseudocode implementation of EnterNotify()
   cyclus::Facility::EnterNotify();
@@ -50,7 +50,7 @@ void FusionPowerPlant::EnterNotify() {
 
 }
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tick() {
   //pseudocode implementation of Tick():
   
@@ -59,22 +59,7 @@ void FusionPowerPlant::Tick() {
     fuel_refill_policy.Start();
     SequesterTritium();
     OperateReactor();
-
-
-    //Maybe make this a function called CycleBlanket() or something.
-    //Leaving it in like this for traceability
-    if (BlanketCycleTime()) {
-      if (blanket->quantity() >= blanket_turnover) {
-        blanket_waste.Push(blanket->ExtractQty(blanket_turnover));
-
-        //guarantee blanket has enough material in CheckOperatingConditions()
-        blanket->Absorb(blanket_feed.Pop(blanket_turnover));
-
-        RecordOperationalInfo("Blanket Cycled");
-      } else {
-        RecordOperationalInfo("Blanket Not Cycled");
-    }
-  }
+    CycleBlanket();
 
   } else {
     //Some way of leaving a record of what is going wrong is helpful info I think
@@ -87,7 +72,7 @@ void FusionPowerPlant::Tick() {
 
 }
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tock() {
   //This is where we used to squash tritium_storage and blanket, but that's no
   //longer needed. Leaving a comment to remind myself about that.
@@ -95,6 +80,22 @@ void FusionPowerPlant::Tock() {
   //Again, not sure about the recording:
   RecordInventories(all_of_them.quantity());
   
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+void FusionPowerPlant::CycleBlanket() {
+  if (BlanketCycleTime()) {
+    if (blanket->quantity() >= blanket_turnover) {
+      blanket_waste.Push(blanket->ExtractQty(blanket_turnover));
+
+      //guarantee blanket has enough material in CheckOperatingConditions()
+      blanket->Absorb(blanket_feed.Pop(blanket_turnover));
+
+      RecordOperationalInfo("Blanket Cycled");
+    } else {
+      RecordOperationalInfo("Blanket Not Cycled");
+    }
+  }
 }
 
 // WARNING! Do not change the following this function!!! This enables your
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -69,10 +69,21 @@ void FusionPowerPlant::Tick() {
   ExtractHelium();
   MoveExcessTritiumToSellBuffer();
 
-  //This maybe belongs in its own function?
+  //Redone blanket implementation:
   if (BlanketCycleTime()) {
-    if (blanket.quantity() >= blanket_turnover) {
-      blanket_excess.Push(blanket.Pop(blanket_turnover));
+    if (blanket->quantity() >= blanket_turnover) {
+      blanket_excess.Push(blanket->ExtractQty(blanket_turnover));
+      
+      //Temporary comment:
+      //Refill the in-core blanket. Blanket storage might be empty, though,
+      //which I think breaks the sim, so I've incuded try/catch to stop that
+      try {
+        blanket->Absorb(blanket_storage.Pop(blanket_turnover));
+      } catch (const std::exception& e) {
+        Record("Blanket Refill Error", e.what());
+        LOG(cyclus::LEV_INFO2, "Reactor") << e.what();
+      }
+
       RecordOperationalInfo("Blanket Cycled");
     } else {
       RecordOperationalInfo("Blanket Not Cycled");
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -11,13 +11,97 @@ std::string FusionPowerPlant::str() {
 }
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+void FusionPowerPlant::EnterNotify() {
+  //Pseudocode implementation of EnterNotify()
+  cyclus::Facility::EnterNotify();
+  fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / seconds_per_year * context()->dt());
+  fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
+  blanket_turnover = blanket_size * blanket_turnover_rate;
+
+  //Truncated for readability. It goes on Tritium Storage and gets started, though.
+  fuel_startup_policy.Init(&tritium_storage).Set(fuel_incommod).Start();
+  
+  //Truncated for readability
+  blanket_fill_policy.Init(&blanket).Set(blanket_incommod).Start();
+
+  //Tritium Buy Policy Section:
+  if (refuel_mode == "schedule") {
+    //Set up a schedule fill policy with Katie's active/dormant policy code
+    //(not shown here). Do not start the policy yet.
+    fuel_refill_policy.Init(&tritium_storage).Set(fuel_incommod);
+  } else if (refuel_mode == "fill") {
+    //otherwise we just do it normally, but still don't start it.
+    fuel_refill_policy.Init(&tritium_storage).Set(fuel_incommod);
+  } else {
+    //This can be more sophisticated later.
+    throw cyclus::KeyError("Refill mode not recognized!");
+  }
+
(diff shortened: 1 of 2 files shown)


src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+  helium_sell_policy.Init(&helium_storage).Set(he3).Start();
+
+}
+
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tick() {
-  std::cout<<"Hello"<<std::endl;
+  //pseudocode implementation of Tick():
+  
+  if (can_operate) {
```
gonuke, line 49: instead of `can_operate` being a Boolean variable, it could be a function that is evaluated each time

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+  
+  if(!can_operate) {
+    try { 
+      CheckOperatingConditions(); //Previously "Startup()"
+      fuel_startup_policy.Stop();
+      fuel_refill_policy.Start();
+    } catch (const std::exception& e) {
+      RecordError("Startup Error", e.what());
+      LOG(cyclus::LEV_INFO2, "FusionPowerPlant") << e.what();
+    }
```
gonuke, line 93: exceptions shouldn't be the way to handle the set of regularly possible states

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+  ExtractHelium(tritium_storage);
+
+  DecayInventory(tritium_excess);
+  ExtractHelium(tritium_excess);
+
+  sequestered_tritium->Decay(context()->time());
+
+  //This was a bunch of free-floating code earlier, but i've truncated it 
+  //because it's mostly for DRE stuff.
+  MoveExcessTritiumToSellBuffer();
```
gonuke, line 65: Agree with initial comment here.  For now we can just rely on pseudo-code:
```suggestion
 
  DecayInventories();
  ExtractHelium();
  MoveExcessTritiumToSellBuffer();
```

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+
+  //This maybe belongs in its own function?
+  if (context()->time() % blanket_turnover_frequency == 0 && !blanket.empty()) {
```
gonuke, line 68: This condition could be it's own function:
```suggestion
  if (blanket_cycle_time()) {
```

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tock() {
-  std::cout<<"There!"<<std::endl;
+  //pseudocode implementation of Tock():
+  
+  if(!can_operate) {
+    try { 
+      CheckOperatingConditions(); //Previously "Startup()"
+      fuel_startup_policy.Stop();
+      fuel_refill_policy.Start();
```
gonuke, line 89: We should discuss whether we can come up with a better buy policy that doesn't require this

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+      RecordError("Startup Error", e.what());
+      LOG(cyclus::LEV_INFO2, "FusionPowerPlant") << e.what();
+    }
+  }
+
+  //I think this behavior got added to cyclus earlier, but am not sure what the
+  //overall functionality of it is at the moment. Now may be a good time to fold
+  //that new behavior in, though, if it's functional. 
+  CombineInventory(tritium_storage);
+  CombineInventory(blanket);
```
gonuke, line 100: Yes - I think new ResBuf features make this unnecessary - see option`is_bulk_`.

src/fusion_power_plant.cc
gonuke: Can you make a comment about what the difference is between these policies?

gonuke: Or maybe a function pseudocode that describes it

gonuke (changes requested):
I'm proposed an even simpler pseudocode in some places

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,98 @@ std::string FusionPowerPlant::str() {
...
+  blanket_fill_policy.Init(&blanket).Set(blanket_incommod).Start();
+
+  //Tritium Buy Policy Section:
+  if (refuel_mode == "schedule") {
+    //Set up a schedule fill policy with Katie's active/dormant policy code
+    //(not shown here). Do not start the policy yet.
+    fuel_refill_policy.Init(&tritium_storage).Set(fuel_incommod);
+  } else if (refuel_mode == "fill") {
+    //otherwise we just do it normally, but still don't start it.
+    fuel_refill_policy.Init(&tritium_storage).Set(fuel_incommod);
```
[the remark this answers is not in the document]
dean-krueger: Yeah, I left the active/dormant stuff out when I truncated them. I'll add that back in, though.

src/fusion_power_plant.cc
dean-krueger: Agreed, though, if I remember correctly there was some issue with simulation-ending errors which I was catching with the try/catch block (I see you have a comment about that later on as well). This is definitely something to try in the full implementation, just mentioning it in case it doesn't fully work.

src/fusion_power_plant.cc
dean-krueger: I think I was doing this for two reasons: I remember there being some sort of simulation-ending error hidden in here that I was getting around with the catch, and two, within that this was a convenient way to let the user know what was going wrong. Definitely worth revisiting, though. I'll change it in the pseudocode implementation and then if we run into problems we can revisit this type of solution (or something better if there is such a thing)

src/fusion_power_plant.cc
dean-krueger: Genuine question: since it's all on one line anyway, does leaving the condition written out like this not improve readability? I agree we could make it a function, but isn't that just making someone go and look at what that one-liner is somewhere else?. Maybe I'm thinking about that wrong, though.

src/fusion_power_plant.cc
dean-krueger: fuel_startup_policy is an (s,S) policy (to use Katie's lingo) which demands the reactor be filled to some user-determined startup fuel inventory whenever it's below that amount (for instance, buy any amount of tritium up to a total inventory of 5kg while below 5kg), whereas the idea with the fuel_refill_policy was that it could be a number of types of buying policies. I implemented two: a "fill" strategy which is basically identical to the startup policy, and a "schedule" policy which buys a fixed amount every few timesteps determined by the user (for instance, once every 6 months it buys 1 kg). The idea was that more could be implemented in the future, since this was intended as a first pass at the agent, but that basically this is where we switch from the buying strategy to get going initially to operating.

src/fusion_power_plant.cc
dean-krueger: Cool, I'll look at that code and see what the best way to represent that built-in functionality with pseudocode.

src/fusion_power_plant.cc
gonuke: While your approach may be more transparent, it increases the cognitive burden.  By making it a function, the reader understands the semantic meaning of the conditions.  If they don't immediately care how that condition is met, they can just read on and understand what's happening.  If they do care how it's met, they can go and look at that function, process what it is doing in isolation, and then set aside the details of how that condition is being met when they read the overall algorithm.

dean-krueger: Ahh, gotcha. Makes sense. Thanks!

dean-krueger:
Made a bunch of the suggested changes. The jury is still out on a better way to do the Start/Stop of those buy policies, but maybe that's a good thing to talk about in the Fuel Cycle meeting this week. I'll add it to the agenda.

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,87 @@ std::string FusionPowerPlant::str() {
...
+ 
+  DecayInventories();
+  ExtractHelium();
+  MoveExcessTritiumToSellBuffer();
+
+  //This maybe belongs in its own function?
+  if (BlanketCycleTime()) {
+    if (blanket.quantity() >= blanket_turnover) {
+      blanket_excess.Push(blanket.Pop(blanket_turnover));
+      CombineInventory(blanket_excess);
```
gonuke, line 76: I think our new `ResBuf` capability means we can skip this

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,87 @@ std::string FusionPowerPlant::str() {
...
+
+ 
+  DecayInventories();
+  ExtractHelium();
+  MoveExcessTritiumToSellBuffer();
+
+  //This maybe belongs in its own function?
+  if (BlanketCycleTime()) {
+    if (blanket.quantity() >= blanket_turnover) {
+      blanket_excess.Push(blanket.Pop(blanket_turnover));
```
gonuke, line 75: When do we refill the blanket?

gonuke (commented):
A couple of final things

src/fusion_power_plant.cc
dean-krueger: Good catch, you're right.

src/fusion_power_plant.cc
gonuke: right - you buy directly into the blanket - there is no blanket inventory before things go into the core?

gonuke: I think it makes sense to have an inventory of fresh blanket material that is not just automatically in the core and is being inserted and removed as appropriate

dean-krueger: So the way it has been working is that in the `OperateReactor()` function, we call some other function called `BreedTritium()`, which pulls the ENTIRE blanket out of that `blanket` resbuf, transmutes some of it into Tritium and He4 (the blanket is required to be Enriched Lithium for this version of the model, but we had plans to change that later on) with the amount being determined by what the user sets as the TBR for the facility, and then we go in and extract the tritium from that composition and then put everything else back into `blanket`. 

The reason I modeled it this way was that I was under the impression that there was just some big tank of liquid enriched lithium and effectively a big swirly pipe that ran through the "blanket" region of the core (maybe that's the wrong word but you hopefully know what I'm talking about), and that the liquid in the tank was being pushed through the pipe fast enough that over the course of a month (standard cyclus timestep) all or most of the liquid would go through the pipe, see some neutrons, and head back into the tank. Then, every once in a while you scoop some "depleted" blanket material out of the tank and pour some "fresh" blanket material in. That may not be the case, though.  

What I hear you saying is that we should have a separate resbuf for fresh blanket material storage which would act as a buffer against limited supply or something (?), and then another resbuf which is the "active" blanket and which has the same role as the current `blanket` resbuf has now but which interacts with the reserve, "fresh" blanket? 

Happy to discuss this more during tomorrow's meeting (or we can push it to Friday if the Fusion crowd is something we want to pull into the conversation) if you think that would be helpful. Hopefully this sort of explains what i was thinking, though.

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,97 @@ std::string FusionPowerPlant::str() {
...
+      
+      //Temporary comment:
+      //Refill the in-core blanket. Blanket storage might be empty, though,
+      //which I think breaks the sim, so I've incuded try/catch to stop that
+      try {
+        blanket->Absorb(blanket_storage.Pop(blanket_turnover));
+      } catch (const std::exception& e) {
+        Record("Blanket Refill Error", e.what());
+        LOG(cyclus::LEV_INFO2, "Reactor") << e.what();
+      }
```
gonuke, line 85: Instead of failing this like we should either refill with what is available or not refill at all, and then the blanket quantity can be part of the condition for operation or not.

If we run out of blanket, we just don't operate.

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,97 @@ std::string FusionPowerPlant::str() {
...
+  }
+
+ 
+  DecayInventories();
+  ExtractHelium();
+  MoveExcessTritiumToSellBuffer();
+
+  //Redone blanket implementation:
+  if (BlanketCycleTime()) {
+    if (blanket->quantity() >= blanket_turnover) {
```
gonuke, line 74: Do you need to update the header to include all the right "in-core" materials?

gonuke (changes requested):
A few more comments not he current pseudo-code

src/fusion_power_plant.cc
dean-krueger: Makes sense to me

src/fusion_power_plant.cc
dean-krueger: Yes. I was leaving them out because there is a lot of minutiae that goes into adding them, and I was worried about bugs, but I can toss them in and just make that future me's problem.

src/fusion_power_plant.h
```diff
@@ -68,6 +68,22 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+  //something like this... Not sure if this will work here though...
+  const cyclus::Composition::Ptr enriched_li = context->GetRecipe(blanket_inrecipe);
+  const cyclus::Composition::Ptr tritium_comp = cyclus::Composition::CreateFromAtom(T);
+
+  //Materials:
+  cyclus::Material::Ptr sequestered_tritium = cyclus::Material::CreateUntracked(0.0, tritium_comp);
+  
+  //This might make more sense to do in EnterNotify()? I'll have to see how it
+  //works outside of pesudocode.
+  cyclus::Material::Ptr blanket = cyclus::Composition::CreateFromAtom(0.0, enriched_li);
```
gonuke, line 85: You probably can't/shouldn't call these functions from the header.  Put them in `EnterNotify()` or something

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,91 @@ std::string FusionPowerPlant::str() {
...
+      if (blanket->quantity() >= blanket_turnover) {
+        blanket_excess.Push(blanket->ExtractQty(blanket_turnover));
+
+        //guarantee blanket has enough material in CheckOperatingConditions()
+        blanket->Absorb(blanket_storage.Pop(blanket_turnover));
+
+        RecordOperationalInfo("Blanket Cycled");
+      } else {
+        RecordOperationalInfo("Blanket Not Cycled");
+    }
```
gonuke, line 76: Maybe this goes in a method called `CycleBlanket()` for readability?

src/fusion_power_plant.cc
```diff
@@ -10,14 +10,91 @@ std::string FusionPowerPlant::str() {
...
+  }
+
+  } else {
+    //Some way of leaving a record of what is going wrong is helpful info I think
+    Record(Error);
```
gonuke, line 81: Standard Cyclus practice is to interrogate the normal Cyclus tables for this rather than being super verbose with new tables.  

At some point, we should discuss where those normal Cyclus tables are inadequate.

gonuke (changes requested):
A few more changes before we move on...

src/fusion_power_plant.cc
dean-krueger: I agree this would be a good conversation. These verbose tables were really helpful for troubleshooting during development, but also when speaking with Ben, we (Matthew Ben and I) came to the conclusion that having a convenient way of letting the user know what error states the facility found itself in over the course of a simulation was a good thing. That may be best to do using the cyclus Log functionality, though, and not with a table.

gonuke (approved):
I definitely think logging is preferred to tables, but we can move this forward now.

Issue #5, opened 2024-08-06
Define the possible states of the fusion power plant


gonuke:
During the course of operation, the fusion power plant can find itself in a finite number of discrete states, based largely on the inventories of its various resource buffers.

Define each of these states and which processes (#4) occur in each state.

PR #11, opened 2024-09-11
Defining Possible States of the Fusion Power Plant

dean-krueger:
Added temporary comments to the code to lay out the general states of the fusion power plant facility. The code below each comment, combined with the comment itself should hopefully be descriptive enough to start the discussion on the different states of the facility.

This PR closes issue #5

### DIFF
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -10,6 +10,10 @@ std::string FusionPowerPlant::str() {
   return Facility::str();
 }
 
+
+//-----------------------------------------------------------//
+//                      Initialization                       //
+//-----------------------------------------------------------//
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::EnterNotify() {
   //Pseudocode implementation of EnterNotify()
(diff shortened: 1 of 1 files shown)


gonuke:
I think this issue was probably poorly defined on my part.  It should probably focus on what causes a change of state since I think the states are largely well defined by #10.

dean-krueger:
Changed the PR around a bit to better match the updated requirements. Also: differentiated between the initial startup condition and the operating condition, since the initial startup condition has that whole bit with the startup inventory requirement being different than the operating inventory requirement.

gonuke:
Sorry for changing the target too much/again... but I think you can start actually coding the functions that you call to do the tests, i.e. moving away from pseudo-code to real code.

dean-krueger:
Sounds good. I'll start with the check functions, but do you also mean that i should do all the other ones (ie `SequesterTritium()`, `OperateReactor()`, etc)?

gonuke:
Just the check functions

dean-krueger:
Going to close this PR for now, since the goals have changed so much. I think this is going to make more sense if we circle back around to it after some other stuff gets merged in (I think that's how this works, at least)

Issue #14, opened 2024-09-15
Add Necessary Code to Header File


dean-krueger:
Within the FusionPowerPlant module there is a lot of structural code that goes in the header file. This code defines user input as well as some values which would otherwise be magic numbers and enables some important functions of the code. Add this code to the repo for review.

Issue #15, opened 2024-09-15
Transition EnterNotify() from pseudocode to real code


dean-krueger:
The EnterNotify() function is the longest function in the FusionPowerPlant, and currently it is all pseudocode. After adding the necessary code to the header file, transition this function to real code for review.

PR #16, opened 2024-09-18
Added necessary elements to header file to support EnterNotify() Function

dean-krueger:
Added all the structure to the header file to support a fully fleshed out EnterNotify(), but did not add that fully fleshed out version of EnterNotify(), since that will be a later PR. This code will also go a long way in supporting the other functions, but I did not add all of that because I figured this would be a better "bite sized" piece to review.

This PR closes Issue #14

### DIFF
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -10,6 +10,10 @@ std::string FusionPowerPlant::str() {
   return Facility::str();
 }
 
+
+//-----------------------------------------------------------//
+//                      Initialization                       //
+//-----------------------------------------------------------//
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::EnterNotify() {
   //Pseudocode implementation of EnterNotify()
@@ -58,18 +62,40 @@ void FusionPowerPlant::EnterNotify() {
 void FusionPowerPlant::Tick() {
   //pseudocode implementation of Tick():
   
+  //-----------------------------------------------------------//
+  //               Check Operational Conditions                //
+  //-----------------------------------------------------------//
   if (CheckOperatingConditions()) {
+
+    //-----------------------------------------------------------//
+    //                Set Operational Buy Policy                 //
+    //-----------------------------------------------------------//
     fuel_startup_policy.Stop();
     fuel_refill_policy.Start();
+
+    //-----------------------------------------------------------//
+    //                  Tritium Sequestration                    //
+    //-----------------------------------------------------------//
     SequesterTritium();
+
+    //-----------------------------------------------------------//
+    //                        Operation                          //
+    //-----------------------------------------------------------//
     OperateReactor();
     CycleBlanket();
 
   } else {
     //Some way of leaving a record of what is going wrong is helpful info I think
+
+    //-----------------------------------------------------------//
+    //                    Failure to Operate                     //
+    //-----------------------------------------------------------//
     Record(Error);
   }
 
+  //-----------------------------------------------------------//
+  //                     Buffer Cleanup                        //
+  //-----------------------------------------------------------//
   DecayInventories();
   ExtractHelium();
   MoveExcessTritiumToSellBuffer();
@@ -82,6 +108,10 @@ void FusionPowerPlant::Tock() {
   //longer needed. Leaving a comment to remind myself about that.
 
   //Again, not sure about the recording:
+
+  //-----------------------------------------------------------//
+  //                      Take Inventory                       //
+  //-----------------------------------------------------------//
   RecordInventories(all_of_them.quantity());
   
 }
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -11,9 +11,9 @@ std::string FusionPowerPlant::str() {
 }
 
 
-//-----------------------------------------------------------//
-//                      Initialization                       //
-//-----------------------------------------------------------//
+//State Boundary
+//Changes from "Unbuilt" to "Newly Built"
+//Conditions: Enter Time = timestep (institution side)
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::EnterNotify() {
   //Pseudocode implementation of EnterNotify()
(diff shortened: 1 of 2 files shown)


src/fusion_power_plant.h
```diff
@@ -47,9 +47,6 @@ class FusionPowerPlant : public cyclus::Facility  {
 
   #pragma cyclus note {"doc": "A stub facility is provided as a skeleton " \
                               "for the design of new facility agents."}
-  //Functions:
-  /// A verbose printer for the FusionPowerPlant
-  virtual std::string str();
```
gonuke, line 52: Does this mean you're just relying on the default `str()` function for `Facility`?  Or do you perhaps want to comment this out and revisit later?

gonuke: I think it's pretty common to implement this to describe the specific details of this kind of facility.

src/fusion_power_plant.h
```diff
@@ -59,27 +56,170 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+  bool CheckInitialStartupCondition();
+  void SequesterTritium();
+  void OperateReactor();
+  void CycleBlanket();
+  void DecayInventories();
+  void ExtractHelium();
+  void MoveExcessTritiumToSellBuffer();
+
+  private:
+    bool initial_startup_condition_met = false;
```
gonuke, line 189: Since you have a function to do this, I don't think you will need to save this state.  Rather, that function will test a variety of conditions each time and return true/false.

src/fusion_power_plant.h
```diff
@@ -59,27 +56,170 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+    double fuel_usage_atoms;
+    double fuel_usage_mass;
+    const double seconds_per_year = 2629846*12;
```
gonuke, line 192: I feel as though we need to make this a constant available in Cyclus for uniformity...  and it should possibly be based on a days
```
seconds_per_year = 365.25 * 24 * 3600
seconds_per_month = seconds_per_year / 12
```

src/fusion_power_plant.h
```diff
@@ -59,27 +56,170 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+    const double MW_to_GW = 1000.0;
+    double avagadros_number = 6.022e23;
```
gonuke, line 194: This should definitely be pulled from elsewhere.  I know it's available to cyclus from pyne, but I don't see it used.  I think most use cases for this are offloaded to pyne, so we'll keep an eye out for that...

src/fusion_power_plant.h
```diff
@@ -59,27 +56,170 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+
+    //NucIDs for Pyne
+    const int tritium_id = 10030000;
+    const int He3_id = 20030000;
+    const int He4_id = 20040000;
+    const int Li6_id = 30060000;
+    const int Li7_id = 30070000;
+
+    //masses in kg/atom
+    const double amu_to_kg = 1.66054e-27;
```
gonuke, line 207: Another quantity we shouldn't really need here

src/fusion_power_plant.cc
```diff
@@ -57,19 +61,43 @@ void FusionPowerPlant::EnterNotify() {
...
+
+
+  //State boundary
+  //Changes from "Newly Built" to "Shut Down"
+  //Conditions: initial_startup_condition_met == True
+  //Requirements:
+  //tritium_storage.quantity() >= startup_inventory_requirement (state variable)
+  //Fuel is of correct composition
+  if (!initial_startup_condition_met) { 
+    CheckInitialStartupCondition()
```
gonuke, line 73: This test should just be part of `CheckOperatingConditions()`.  Not sure why we need to keep track of the `initial_startup_condition_met`  if we can always query it with `CheckInitialStartupCondtion()`?

gonuke (changes requested):
Thanks for leaping in here.  A few questions about physical constants and the change to the state logic.

src/fusion_power_plant.h
dean-krueger: I removed this because of a conversation i thought we had on the original PR about this function not being really used for anything. I can add it back, though, especially since I can't find that conversation anymore...

src/fusion_power_plant.h
dean-krueger: So this one's actually a little tricky. Because the initial startup inventory condition is unique and only needs to be met once at the beginning of operation, I believe we do in fact have to store that it was met. If we continue to test for it after that initial startup it should (rightly) fail. This has to do with the startup inventory being greater than the operational inventory (the initial sequestration takes a big hit on the on hand tritium on the first timestep). We want to make sure that on that first operational timestep we have `startup_inventory` amount of tritium, but after that we only need to make sure that we have `reserve_inventory` amount, or more importantly enough on hand to operate for a full timestep.

src/fusion_power_plant.h
dean-krueger: This is a whole rabbit hole. Cyclus seems to use some variety of rounded Gaussian year which is ~365.256 days long, and I've never fully understood why that is, or what the rounding convention was (it's not exactly that which is just so confusing to me). This is defined in context.h as `const uint64_t kDefaultTimeStepDur = 2629846` with no comments saying why that is. Happy to discuss further.

src/fusion_power_plant.h
dean-krueger: Sounds good. Also changed it so that the spelling of Avogadro was correct.

src/fusion_power_plant.h
dean-krueger: Because we should get it from pyne?

src/fusion_power_plant.cc
dean-krueger: See my earlier comment, but just to summarize: we need to meet a specific "startup condition" to start operating the first time, but then shouldn't need to meet that condition ever again, but it's important always that we've met that condition that one time.

src/fusion_power_plant.h
gonuke: Yeah - there is probably a good reason to clarify/fix this for all of Cyclus - we should launch a convo in an issue and/or slack to get Katie & Amanda's input (among others)

src/fusion_power_plant.h
gonuke: Or avoid using it at all by using PyNE's ability to translate from mass to atoms (maybe?)

src/fusion_power_plant.cc
gonuke: What happens if we become completely depleted of tritium, do we enter this state again? Is that possible?

dean-krueger: The reason for the different startup condition is to initially saturate all the parts on the plant with tritium (so far as I understand it, at least), so if the plant runs fully out of tritium during operation we shouldn't need to re-enter this state since presumably the plant is already fully, or close to fully, saturated. I can imagine an edge case where a plant would go offline for like 30 years and then maybe need to re-saturate since a significant amount of the saturated tritium has decayed, but a. that feels like a whole different can of worms from the "normal" envisioned use case of this, and b. I feel like, similar to mothballing a fission plant for 30 years, you wouldn't just be trying to flip the switch on it coming back up again and sticking fresh fuel into the facility. That'd be a whole different refurbishment operation which I think is maybe outside the scope of what we're trying to model. 

Thus, we shouldn't ever need to re-enter this state again, and therefore once the facility moves out of this state, there is no way to move back into it.

src/fusion_power_plant.h
dean-krueger: I was unaware of this capability (I don't typically use pyne). I'll leave this in here for now so things can get up and running, but make a comment and a to-do item to try to switch to using some pyne features to do this type of operation instead, if that works for you?

src/fusion_power_plant.h
dean-krueger: I can do both now!

src/fusion_power_plant.cc
gonuke: So can't we use the sequestered tritium inventory as a flag for this?

dean-krueger: we can. I will do that.

src/fusion_power_plant.h
gonuke: It's not strictly necessary, but it is useful to have a way to print out information about a prototype/facility and its configuration parameters.

src/fusion_power_plant.h
```diff
@@ -59,27 +56,173 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+    const int Li7_id = 30070000;
+
+    //masses in kg/atom
+    //Maybe make pyne do some of this...
+    const double amu_to_kg = 1.66054e-27;
+    const double Li7_atomic_mass = pyne::atomic_mass(Li7_id) * amu_to_kg;
+    const double Li6_atomic_mass = pyne::atomic_mass(Li6_id) * amu_to_kg;
+    const double tritium_atomic_mass = pyne::atomic_mass(tritium_id) 
+                 * amu_to_kg;
+    const double He4_atomic_mass = pyne::atomic_mass(He4_id) * amu_to_kg;
```
gonuke, line 215: Arguably, we don't need to define these until we are sure what we need as you write the algorithms elsewhere...  This is also what's driving the need for `amu_to_kg`

gonuke (changes requested):
Upon further review, this PR may be adding more than we need....  One of the reasons to break things up is to only add variables to the class definition once you are sure they are needed, largely because you've added algorithms/methods that need them. 

It makes it hard to know if these are the right variables when we don't know if/how they will be used.

src/fusion_power_plant.h
dean-krueger: Removed a lot of the code that was non-essential. Should I go ahead and fold Issue #15 into this PR so that you can see what (a trimmed down) EnterNotify() is doing, or should we leave that for the next one? I forget what we ultimately decided yesterday.

gonuke: I am interested in seeing the PR associated with #15 as it's own PR into `main` to see what it requires.

src/fusion_power_plant.cc
gonuke: Were you still planning to make this change, @dean-krueger ?

PR #17, opened 2024-10-02
Change EnterNotify from pesudocode to real code

dean-krueger:
This PR closes issue #15 

Rewrote EnterNotify() to change pseudocode into real code. This PR should also help clarify PR #16


src/fusion_power_plant.cc
```diff
@@ -12,45 +12,71 @@ std::string FusionPowerPlant::str() {
...
+  //blanket_turnover = blanket_size * blanket_turnover_rate;
 
-  //Truncated for readability. It goes on Tritium Storage and gets started, though.
-  fuel_startup_policy.Init(&tritium_storage).Set(fuel_incommod).Start();
-  
-  //Truncated for readability
-  blanket_fill_policy.Init(&blanket).Set(blanket_incommod).Start();
+  //Create the blanket material for use in the core, no idea if this works...
+  const Composition::Ptr enriched_li = context->GetRecipe(blanket_inrecipe);
+  Material::Ptr blanket = Composition::CreateFromAtom(0.0, enriched_li);
```
gonuke, line 24: This may work, but this variable will be in local scope and expire when the function ends.  I think `blanket` needs to be a member variable of the class.

src/fusion_power_plant.cc
```diff
@@ -12,45 +12,71 @@ std::string FusionPowerPlant::str() {
...
-  fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
-  blanket_turnover = blanket_size * blanket_turnover_rate;
 
-  //Create the blanket material for use in the core
-  const cyclus::Composition::Ptr enriched_li = context->GetRecipe(blanket_inrecipe);
-  cyclus::Material::Ptr blanket = cyclus::Composition::CreateFromAtom(0.0, enriched_li);
+  //fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / 
+  //  seconds_per_year * context()->dt());
+  //fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
+  //blanket_turnover = blanket_size * blanket_turnover_rate;
```
gonuke, line 20: I'm assuming that we're just commenting these out until we actually get to a method that needs to use these variables.

gonuke (changes requested):
A few small comments here, but also, feel free to add all the member variables to the header that are now explicitly necessary, as part of this PR (extracted from #16) as necessary

src/fusion_power_plant.cc
dean-krueger: Yeah, I thought about deleting them, but figured since we'll likely add them all back in the next step it was fine to just leave them as comments.

src/fusion_power_plant.cc
dean-krueger: I think you're right. There was a bit of confusion between the comment that said to move these to here and this (especially since none of this builds anyway, so it's hard to tell what's going to be an issue). I'll add blanket as a member variable, then just set its value here in EnterNotify, similar to how fuel_usage_mass and those other commented out variables work.

dean-krueger:
This PR will also close #14 now that I've added all the header file code.

src/fusion_power_plant.cc
```diff
@@ -12,45 +12,71 @@ std::string FusionPowerPlant::str() {
...
 
-  //Truncated for readability. It goes on Tritium Storage and gets started, though.
-  fuel_startup_policy.Init(&tritium_storage).Set(fuel_incommod).Start();
-  
-  //Truncated for readability
-  blanket_fill_policy.Init(&blanket).Set(blanket_incommod).Start();
+  //Create the blanket material for use in the core, no idea if this works...
+  const Composition::Ptr enriched_li = context->GetRecipe(blanket_inrecipe);
+  blanket = Composition::CreateFromAtom(0.0, enriched_li);
```
gonuke, line 24: We can probably just combine these unless we'll need this composition elsewhere.

Also - `blanket` is a `Material::Ptr` and `Composition::CreateFromAtom()` returns a `Composition::Ptr` - might need:
```suggestion
  blanket = Material::Create(this, 0.0, Composition::CreateFromAtom(0.0, context->GetRecipe(blanket_inrecipe)));
```

src/fusion_power_plant.h
```diff
@@ -59,27 +56,103 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+  std::string blanket_inrecipe;
+
+  #pragma cyclus var { \
+    "doc": "Blanket waste commodity designation", \
+    "tooltip": "Blanket waste commodity", \
+    "uilabel": "Blanket waste commodity" \
+  }
+  std::string blanket_outcommod;
+
+  //Resource Buffers and Trackers:
```
gonuke, line 112: These `ResBuf`s can maybe be private members?

src/fusion_power_plant.h
```diff
@@ -59,27 +56,103 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+  cyclus::toolkit::MatlBuyPolicy fuel_startup_policy;
+  cyclus::toolkit::MatlBuyPolicy fuel_refill_policy;
+  cyclus::toolkit::MatlBuyPolicy blanket_fill_policy;
+
+  cyclus::toolkit::MatlSellPolicy tritium_sell_policy;
+  cyclus::toolkit::MatlSellPolicy helium_sell_policy;
+  cyclus::toolkit::MatlSellPolicy blanket_waste_sell_policy;
+
+  cyclus::toolkit::TotalInvTracker fuel_tracker;
+  cyclus::toolkit::TotalInvTracker blanket_tracker;
```
gonuke, line 128: pretty sure these can all be private members

src/fusion_power_plant.cc
```diff
@@ -12,45 +12,71 @@ std::string FusionPowerPlant::str() {
...
+  // Tritium Buy Policy Selection:
   if (refuel_mode == "schedule") {
-    IntDistribution::Ptr active_dist =
-        FixedIntDist::Ptr(new FixedIntDist(1));
-    IntDistribution::Ptr dormant_dist =
-        FixedIntDist::Ptr(new FixedIntDist(buy_frequency - 1));
-    DoubleDistribution::Ptr size_dist =
-        FixedDoubleDist::Ptr(new FixedDoubleDist(1));
-    //Do not start the policy yet.
+    IntDistribution::Ptr active_dist =
```
gonuke, line 40: You probably need to add some header to have access to these.

gonuke (changes requested):
I think this is close.  Do you think we can get a version of this that will build successfully, even if it does nothing?  e.g. all the functions can just be empty and do nothing.

src/fusion_power_plant.cc
dean-krueger: I'll double check that and make the necessary changes

dean-krueger:
I'll clean up some of these comments while working towards a version of this that builds. Hopefully making this buildable will clarify some of the other issues.

src/fusion_power_plant.h
```diff
@@ -59,19 +63,148 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+  #pragma cyclus var { \
+    "default": 0.05, \
+    "doc": "Percent of blanket that gets recycled every blanket turnover period", \
+    "tooltip": "Defaults to 0.05 (5%), must be between 0 and 15%", \
+    "units": "dimensionless", \
+    "uitype": "range", \
+    "range": [0, 0.15], \
+    "uilabel": "Blanket Turnover Rate" \
+  }
+  double blanket_turnover_quantity;
```
gonuke, line 157:
```suggestion
  double blanket_turnover_fraction;
```

gonuke (changes requested):
Just one variable name change recommendation and then this should be good to go.

PR #21, opened 2024-10-30
Test and Debug PR

### LINKED ISSUE #22: Testing and Debugging required
Tricycle is filled out, but it still needs to be checked for bugs and to make sure it's doing what we want/expect it to do.

dean-krueger:
Changes to FusionPowerPlant which came out of some basic testing and debugging. This PR does not address the lack of formal testing, but rather gets the code doing more or less what we expect it to do. 

This PR Closes #22


src/fusion_power_plant.cc
```diff
@@ -106,25 +110,35 @@ void FusionPowerPlant::EnterNotify() {
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tick() {
-  //pseudocode implementation of Tick():
+
+  // For Debugging:
+  std::cout<<"Timestep "<<context()->time()<<std::endl;
```
gonuke, line 115: Even for debugging, it would be good to look at using logging

gonuke (changes requested):
A comment on logging

src/fusion_power_plant.cc
```diff
@@ -227,9 +259,9 @@ void FusionPowerPlant::OperateReactor() {
 }
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::DecayInventories() {
-  //Left empty to quickly check if code builds
-  //tritium_storage.Decay();
-  //tritium_excess.Decay();
+  tritium_storage.Decay();
+  tritium_excess.Decay();
+  sequestered_tritium->Decay(context()->time());
```
gonuke, line 264: I think this does the current time by default, just like the resbuf Decay, so you could just call `Decay()` with no arguments here to be like the others.

figures/TricycleFPP_Internal_Inv.png
gonuke, line 1: It's nice to see this, but not sure we need it in the repo?  You could include it in the conversation about the PR and also comment on what we should take away from this.

src/fusion_power_plant.cc
```diff
@@ -34,28 +34,32 @@ std::string FusionPowerPlant::str() {
...
   fuel_startup_policy
-      .Init(this, &tritium_storage, std::string("Tritium Storage"),
-            &fuel_tracker)
-      .Set(fuel_incommod, tritium_comp)
-      .Start();
+    .Init(this, &tritium_storage, std::string("Tritium Storage"),
+          &fuel_tracker, std::string("ss"),
+          startup_inventory,
+          startup_inventory)
+    .Set(fuel_incommod)
```
gonuke, line 57: is there a reason you no longer include the tritium composition?

src/fusion_power_plant.cc
```diff
@@ -34,28 +34,32 @@ std::string FusionPowerPlant::str() {
...
   fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / 
     (kDefaultTimeStepDur * 12) * context()->dt());
-  //fuel_usage_atoms = fuel_usage_mass / tritium_atomic_mass;
   blanket_turnover = blanket_size * blanket_turnover_fraction;
-
+  double startup_inventory = reserve_inventory + sequestered_equilibrium; 
+  
   //Create the blanket material for use in the core, no idea if this works...
-  blanket = Material::Create(this, 0.0, 
+   blanket = Material::Create(this, 0.0,
```
gonuke, line 49: wrong indentation - may use something like `clang-format` to ensure consistent formatting.

gonuke (changes requested):
This looks pretty much ready to go - just a few nit-picking comments.

src/fusion_power_plant.cc
dean-krueger: Unfortunately, it does not. The material version of decay requires a time `Decay (int current_time) {` whereas the ResBuf version requires the same time, but with a default `Decay (int current_time = -1) { `. If you take out the `context()->time()` bit you get an error: `decay cannot use default time with NULL context`.

gonuke: I see... that's because `sequestered_tritium` is no tracked.

gonuke (approved):
This looks good to move forward.

PR #25, opened 2024-11-18
Fixed Various Startup Issues

### LINKED ISSUE #24: Various startup-related issues
These should be two issues, but I'm combining them into one since the second is an extension of the first.

Issue 1:
If FPP gets enough fuel to start operations before it reaches the amount required to officially start up (reserve + sequestered_equilibrium), it will start operating. This circumvents the concept of a required startup inventory in a way that was not intended.

Fix: Add a catch to ReadyToOperate to check that returns false if tritium_storage.quantity < startup_inventory and sequestered_tritium < eps.

Issue 2:
After requiring that FPPs reach startup inventory before operating, it was revealed that the fuel_startup_policy was buying too little fuel in tritium-limited simulations due to the effects of decay. 

Fix: Have FPP's fuel_startup_policy aim to overbuy tritium by roughly (1+tritium decay per tilmestep).

dean-krueger:
This PR closes #24 

As described in the issue, FPPs could behave incorrectly at startup, either starting before they were supposed to (lack of a check for startup-inventory being reached) or would never startup at all in tritium-limited sims.

### DIFF
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -199,14 +199,14 @@ bool FusionPowerPlant::TritiumStorageClean() {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 bool FusionPowerPlant::ReadyToOperate() {
 
-  // determine required tritium storage inventory
-  double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
-
-  // ensure correct startup behavior
-  if (sequestered_tritium->quantity() < cyclus::eps_rsrc() && tritium_storage.quantity() < startup_inventory) {
-    return false;
+  // Determine tritium inventory required to operate
+  if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
+    double required_storage_inventory = reserve_inventory + SequesteredTritiumGap();
+  } else {
+    double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
   }
-  // check basic tritium storage quantity requirement
+
+  // check  tritium storage quantity requirement
   if (tritium_storage.quantity() < required_storage_inventory || !TritiumStorageClean()) {
     return false;
   }
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -13,8 +13,6 @@ using cyclus::toolkit::ResBuf;
 namespace tricycle {
 
 const double FusionPowerPlant::burn_rate = 55.8;
-const double FusionPowerPlant::tritium_yearly_decay_fraction = 0.055;
-
 const double MW_to_GW = 1000;
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
@@ -27,15 +25,12 @@ FusionPowerPlant::FusionPowerPlant(cyclus::Context* ctx) : cyclus::Facility(ctx)
   helium_excess = ResBuf<Material>(true);
   blanket_feed = ResBuf<Material>(true);
   blanket_waste = ResBuf<Material>(true);
-
 }
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 std::string FusionPowerPlant::str() {
   return Facility::str();
 }
-
-
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::EnterNotify() {
@@ -45,11 +40,6 @@ void FusionPowerPlant::EnterNotify() {
     (kDefaultTimeStepDur * 12) * context()->dt());
   blanket_turnover = blanket_size * blanket_turnover_fraction;
   startup_inventory = reserve_inventory + sequestered_equilibrium; 
-
-  // Otherwise tritium-limited sims will never startup. We need to overbuy at
-  // startup.
-  trituim_overbuy_multiplier = 1 + (tritium_yearly_decay_fraction / 
-    (kDefaultTimeStepDur * 12) * context()->dt());
   
   //Create the blanket material for use in the core, no idea if this works...
   blanket = Material::Create(this, 0.0, 
@@ -58,8 +48,8 @@ void FusionPowerPlant::EnterNotify() {
   fuel_startup_policy
     .Init(this, &tritium_storage, std::string("Tritium Storage"),
           &fuel_tracker, std::string("ss"),
-          startup_inventory * trituim_overbuy_multiplier,
-          startup_inventory * trituim_overbuy_multiplier)
+          startup_inventory,
+          startup_inventory)
     .Set(fuel_incommod, tritium_comp)
     .Start();
 
@@ -109,7 +99,6 @@ void FusionPowerPlant::EnterNotify() {
       .Init(this, &blanket_waste, std::string("Blanket Waste"))
       .Set(blanket_outcommod)
       .Start();
-
 }
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
@@ -126,7 +115,6 @@ void FusionPowerPlant::Tick() {
     // Some way of leaving a record of what is going wrong is helpful info I think
     // Use the cyclus logger
   }
-
   
   DecayInventories();
   ExtractHelium();
@@ -144,7 +132,6 @@ void FusionPowerPlant::Tick() {
     fuel_startup_policy.Stop();
     fuel_refill_policy.Start();
   }
-
 }
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
@@ -155,7 +142,6 @@ void FusionPowerPlant::Tock() {
   RecordInventories(tritium_storage.quantity(), tritium_excess.quantity(), 
                     sequestered_tritium->quantity(), blanket_feed.quantity(),
                     blanket_waste.quantity(), helium_excess.quantity());
-
 }
 
 void FusionPowerPlant::RecordInventories(double tritium_storage, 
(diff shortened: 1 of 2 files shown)


src/fusion_power_plant.cc
```diff
@@ -196,9 +201,13 @@ bool FusionPowerPlant::ReadyToOperate() {
...
   double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
 
-  // check tritium storage quantity
+  // check basic tritium storage quantity requirement
   if (tritium_storage.quantity() < required_storage_inventory || !TritiumStorageClean()) {
-        return false;
+    return false;
+  }
+  // insure correct startup behavior
+  if (tritium_storage.quantity() < startup_inventory && sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
```
gonuke, line 209: Maybe this should be first, and perhaps change the order of the conditions.  The order doesn't matter for logic reasons, but communicates the purpose - we only care about this condition when the sequestered tritium is essentially 0

src/fusion_power_plant.cc
```diff
@@ -52,8 +57,8 @@ void FusionPowerPlant::EnterNotify() {
   fuel_startup_policy
     .Init(this, &tritium_storage, std::string("Tritium Storage"),
           &fuel_tracker, std::string("ss"),
-          startup_inventory,
-          startup_inventory)
+          startup_inventory * (1 + trituim_overbuy_multiplier),
```
gonuke, line 60: Should we also apply this correction to the reserve inventory?  Or instead of introducing this at all, could we just expect the user to specify a larger reserve inventory to account for this?

src/fusion_power_plant.h
```diff
@@ -259,6 +259,9 @@ class FusionPowerPlant : public cyclus::Facility  {
   double blanket_turnover;
   double fuel_usage_mass;
   static const double burn_rate; // kg/GW-y
+  double startup_inventory;
+  double tritium_yearly_decay_fraction = 0.055;
```
gonuke, line 263: Should this be a `static const` like the burn rate?

src/fusion_power_plant.cc
```diff
@@ -196,9 +201,13 @@ bool FusionPowerPlant::ReadyToOperate() {
   // determine required tritium storage inventory
   double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
 
-  // check tritium storage quantity
+  // check basic tritium storage quantity requirement
   if (tritium_storage.quantity() < required_storage_inventory || !TritiumStorageClean()) {
-        return false;
+    return false;
+  }
+  // insure correct startup behavior
```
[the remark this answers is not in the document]
dean-krueger: Funny story: I had "ensure" here at first, but I'm bad at spelling, so I googled it and the first thing that came up was the dietary drink thing, so I assumed it was the other one. Oops! I'll change it back. Thanks for the catch.

src/fusion_power_plant.h
dean-krueger: Probably, yeah.

src/fusion_power_plant.cc
dean-krueger: Sure, that makes sense to me.

src/fusion_power_plant.cc
```diff
@@ -196,9 +202,13 @@ bool FusionPowerPlant::ReadyToOperate() {
   // determine required tritium storage inventory
   double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
```
gonuke, line 203: I'm trying to remember why we need `reserve_inventory` + `sequestered_inventory` to startup but only need `fuel_usage_mass + sequesteredTritiumGap` to keep operating.  That is, given some of our back & forth discussion of these variables, why is startup based on `reserve_inventory` and continued operation based only on `fuel_usage_mass`?

src/fusion_power_plant.h
```diff
@@ -259,6 +259,9 @@ class FusionPowerPlant : public cyclus::Facility  {
   double blanket_turnover;
   double fuel_usage_mass;
   static const double burn_rate; // kg/GW-y
+  double startup_inventory;
+  static const double tritium_yearly_decay_fraction;
```
gonuke, line 263: From a readability point of view (and possibly a minor, in the weeds, compilation benefit) it probably makes sense to put the `static const` variables together and as the first (or last??) variables in the class.

src/fusion_power_plant.cc
gonuke: More generally, `SequesteredTritiumGap()` will be the same as `sequestered_equilibrium` before startup.  As such, there may be a path to simplify all of this and not need to store all of `startup_inventory`, `reserve_inventory`, and `fuel_usage_mass`.

gonuke (changes requested):
A few comments about the different inventory variables we save.

src/fusion_power_plant.cc
dean-krueger: The bigger issue was that reserve_inventory was never considered when starting up. Reserve inventory is (supposedly) necessary to start up because plants want to have some certainty that something in their tritium recovery system isn't going to go wrong and cause them to suddenly shut down. However, once the plant is already operating, it doesn't make sense to stop operating simply because they had to dip into their reserves (that's the whole point of the reserve). After startup they only check that they have enough to run for a full time step, but the whole time they're trying to buy back up to their reserve again. I've put this as a discussion point for today's fuel cycle meeting so we can have a more in-depth conversation about this if needed.

src/fusion_power_plant.cc
```diff
@@ -193,12 +199,16 @@ bool FusionPowerPlant::TritiumStorageClean() {
...
 bool FusionPowerPlant::ReadyToOperate() {
 
-  // determine required tritium storage inventory
-  double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
+  // Determine tritium inventory required to operate
+  if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
+    double required_storage_inventory = reserve_inventory + SequesteredTritiumGap();
+  } else {
+    double required_storage_inventory = fuel_usage_mass + SequesteredTritiumGap();
+  }
```
gonuke, line 207: Because of limitations of C++ scope this may need to declare the variable outside of the if statement:
```suggestion
  // Determine tritium inventory required to operate
  double required_storage_inventory = SequesteredTritiumGap();
  if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
    required_storage_inventory += reserve_inventory;
  } else {
    required_storage_inventory += fuel_usage_mass;
  }
```

src/fusion_power_plant.cc
```diff
@@ -43,7 +44,12 @@ void FusionPowerPlant::EnterNotify() {
   fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / 
     (kDefaultTimeStepDur * 12) * context()->dt());
   blanket_turnover = blanket_size * blanket_turnover_fraction;
-  double startup_inventory = reserve_inventory + sequestered_equilibrium; 
+  startup_inventory = reserve_inventory + sequestered_equilibrium; 
+
+  // Otherwise tritium-limited sims will never startup. We need to overbuy at
+  // startup.
+  trituim_overbuy_multiplier = 1 + (tritium_yearly_decay_fraction /
```
gonuke, line 51: 1. you have a typo in `tritium`
2. this is only used in this function, does it need to be a member variable?
3. I think the name of this variable makes is hard to understand/remember why it exists. I wonder if we should instead defined `tritium_timestep_decay_fraction` and multply by `(1 + tritium_timestep_decay_fraction)` where necessary.

src/fusion_power_plant.cc
```diff
@@ -43,7 +44,12 @@ void FusionPowerPlant::EnterNotify() {
   fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) / 
     (kDefaultTimeStepDur * 12) * context()->dt());
   blanket_turnover = blanket_size * blanket_turnover_fraction;
-  double startup_inventory = reserve_inventory + sequestered_equilibrium; 
+  startup_inventory = reserve_inventory + sequestered_equilibrium;
```
gonuke, line 47: 1. Does  `startup_inventory` need to be a member variable if it is only used here?
2. If we define it after `tritium_overbuy_multiplier` (or whatever we end up calling it), can't we just define `startup_inventory` to include that effect rather than multiplying in the initializer for the buy policy?  That is, isn't the actual startup inventory inclusive of the decay effect?

gonuke (changes requested):
A few more little changes to make sure we can understand all the details in the future.

src/fusion_power_plant.cc
dean-krueger: I think I'm going to try to remove this whole thing with the over-purchasing of tritium, honestly. It's kind of confusing, and makes the code do things that aren't obvious or user-defined.

src/fusion_power_plant.cc
dean-krueger: See previous comment about removing this. I think it ended up feeling too kludgey...

src/fusion_power_plant.cc
dean-krueger: I think this has been fixed.

src/fusion_power_plant.h
```diff
@@ -258,7 +258,7 @@ class FusionPowerPlant : public cyclus::Facility  {
   Material::Ptr blanket;
   double blanket_turnover;
   double fuel_usage_mass;
-  static const double burn_rate; // kg/GW-y
+  double startup_inventory;
```
gonuke, line 261: see above note about whether this needs to be a member variable.

gonuke (changes requested):
I think this is nearly finished - just a few little comments.

dean-krueger:
Good point, I was so focused on the larger problem that this one fell through the cracks. Should be fixed now!

gonuke (approved):
This looks good to me.

PR #26, opened 2024-11-19
Add Basic Unit Tests

### LINKED ISSUE #23: Formal Testing Required
FusionPowerPlant needs to have formal unit tests written to proceed.

dean-krueger:
This PR closes #23 

Tricycle lacks formal unit testing. This PR adds the most basic version of those (though more may be warranted). All tests pass except one (`GoogleTestVerification.UninstantiatedParameterizedTestSuite<InstitutionTests>`)  which isn't one of the ones I wrote, and I'm not certain what why it's failing. Either way, felt like this was in a fine place to go forward.


gonuke:
This looks promising.  We should add something to the workflow to run the tests...

dean-krueger:
Unit testing workflow bit added, the failing test is the one I mentioned in the original PR description. I've made discussing this an agenda item for today's fuel cycles meeting, but the workflow does seem to be working.

Only question is: I made the build, and then build and test two separate jobs (it seemed like it would be more obvious what was going on at a glance), but I could easily see the argument that they should be the same job with just two different steps. Interested in your thoughts.

gonuke:
I fear it's going to take a little sleuthing to get to the bottom of this.

dean-krueger:
I can do some sleuthing later this week. It'll be a nice break from the holiday insanity. 

Also, left a comment in PR #24 about this, but some of these unit tests are going to be outdated since I removed the silly tritium overbuy thing I was doing, but they'll be quick fixes (thank you past Dean) and I'll just do them once that PR is merged so I don't have to do everything twice (or worse, worry about branch management)

gonuke:
I'm a little stumped by this and wonder if we need to change the stub in someway based on updates to our build infrastructure.  We may want to get some perspective from @bennibbelink given some of the changes he's made.

gonuke:
My instincts tell me that we are trying to run some tests that are not defined, and the trick will be to find either why we are trying to run them at all or why they are not defined.  My instincts further tell me that we probably want to avoid trying to run them in this instance rather than need to define them.

What my instincts don't tell me is how to figure this out, so some sleuthing is definitely required.

bennibbelink:
I remember running into a similar issue when I added parameterized tests to cyclus for multithreading https://github.com/cyclus/cyclus/issues/1709. I will try to take a closer look at this in the coming days, but it might not be until early next week.  If you are planning on sleuthing earlier than that @dean-krueger that PR might be a good reference

bennibbelink:
I think I have a handle on what is happening here.  Since we link against `libbaseagentunittests.so` GTest sees all the test definitions in that library (`cyclus/tests/agent_tests/*_tests.cc`).  If you run `tricycle_unit_tests --gtest_list_tests` you see not only the tests that @dean-krueger has written in tricycle, but also `PositionFacTest` and the uninstantiated `InstitutionTests`.  I was a bit puzzled why you could not see `RegionTest`, which led me to look at `region_tests.cc` which includes a line at the top `GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(RegionTests);`.  This is the workaround that GTest suggests in the error message, but I wasn't sure if we could link against the testing library in such a way that this wasn't necessary (I haven't found a way to do this without splitting up the base agent tests into separate libraries).  With this as background... there are a few options.

- Add `GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(<suite name>)` to all upstream test suites so that archetype developers don't run into these errors when they only use a subset of the available test suites
- Make it the responsibility of the archetype developer to add `GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(<suite name>)` to their test files for each suite that they do not use
- Restructure the build so that these base agent tests are each their own library, so that an archetype developer can link to the libraries in a granular fashion. (I have not tested this)

I think the first bullet probably makes the most sense, but I'm curious if y'all have any thoughts.

bennibbelink:
@dean-krueger some other comments on your PR:

- `INSTANTIATE_TEST_CASE_P` is deprecated in favor of `INSTANTIATE_TEST_SUITE_P`, so if you change both occurrences you should stop seeing those build warnings.
- You might be able to replace some of the CMake boilerplate with macros from https://github.com/cyclus/cyclus/blob/main/cmake/CyclusBuildSetup.cmake (this cleanup could probably be a separate PR from this one)

gonuke:
It looks like the changes in `region_tests.cc` came in cyclus/cyclus#1738 by @bennibbelink, in response to some change in GTEST, and didn't propagate to `instittion_tests.cc` - probably an oversight and perhaps something I should have caught in review.

gonuke:
Which is to say... I think option 1 is the right choice

gonuke:
I've PR'ed a fix

dean-krueger:
This should be good to go now. Thanks Ben and Paul for the quick fix on that failing test!

src/fusion_power_plant_tests.cc
```diff
@@ -37,18 +64,442 @@ TEST_F(FusionPowerPlantTest, Print) {
...
+      "  <TBR>1.00</TBR>"
+      " <reserve_inventory>6.0</reserve_inventory>"
+      " <sequestered_equilibrium>2.121</sequestered_equilibrium>"
+      " <fuel_incommod>Tritium</fuel_incommod>"
+      " <blanket_incommod>Enriched_Lithium</blanket_incommod>"
+      " <blanket_outcommod>Depleted_Lithium</blanket_outcommod>"
+      " <blanket_inrecipe>enriched_lithium</blanket_inrecipe>"
+      " <blanket_size>1000</blanket_size>"
+      " <blanket_turnover_fraction>0.03</blanket_turnover_fraction>"
+      " <he3_outcommod>Helium_3</he3_outcommod>";
```
gonuke, line 90: All of your sims use essentially this config - maybe it can be defined once globally?  If you want to include the case with the wrong fuel, set the `fuel_incommod` to be `Fuel` and define the source correctly (with tritium) in most cases and incorrectly (with enriched lithium) in that case.

You can also append the extra line for `Li7_contribution` in those cases.

src/fusion_power_plant_tests.cc
```diff
@@ -37,18 +64,442 @@ TEST_F(FusionPowerPlantTest, Print) {
...
+
+  double expected_decay = (init_quant - sequestered) 
+              - (init_quant - sequestered) * std::pow(2, -lambda * t);
+
+  EXPECT_NEAR(expected_decay, he3, 1e-6);
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(FusionPowerPlantTest, Li7EdgeCases) {
+  // Test that FPP does the same thing regardless of Li-7 Contribution
```
gonuke, line 237: Should something be different this this case?  If there is no difference ever, why include it at all?

src/fusion_power_plant_tests.cc
```diff
@@ -37,18 +64,442 @@ TEST_F(FusionPowerPlantTest, Print) {
...
+      " <he3_outcommod>Helium_3</he3_outcommod>";
+
+  int simdur = 4;
+  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:FusionPowerPlant"), config, simdur);
+
+  sim.AddRecipe("tritium", tritium());
+  sim.AddRecipe("enriched_lithium", enriched_lithium());
+
+  sim.AddSource("Tritium").recipe("tritium").Finalize();
+  sim.AddSource("Enriched_Lithium").recipe("enriched_lithium").Finalize();
```
gonuke, line 99: There may be some value in also bundling all of this into a function since most of it happens the same with each test.  At least the following:
* setup the MockSim
* add two recipes
* add one source for enriched lithium

gonuke (changes requested):
Generally this looks like pretty good coverage - I've made a few suggestions for reuse as much as possible.

src/fusion_power_plant_tests.cc
dean-krueger: I went in and did this for all the "standard" sims, but there were two or three that had non-standard implementations so I just left those alone instead of trying to be clever.

src/fusion_power_plant_tests.cc
dean-krueger: I did this in sort of a "dumb" way to avoid the overhead that comes with trying to be clever. I think it achieves the spirit of what you were asking for, but let me know if you want me to go the extra step.

src/fusion_power_plant_tests.cc
```diff
@@ -37,18 +64,442 @@ TEST_F(FusionPowerPlantTest, Print) {
   // Test FusionPowerPlant specific aspects of the print method here
 }
 
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-TEST_F(FusionPowerPlantTest, Tick) {
```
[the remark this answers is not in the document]
dean-krueger: This test is just the default generated by the stub, and it checks that there is no error thrown in the tick function, but it runs it without context, so it gets an error from when we peak at the contents of tritium_storage during the ReadyToOperate check since there's nothing in there. This doesn't affect real sims when it's run, but it does make this test always fail, so I removed it since it didn't seem like this broke the code and we test the functionality of Tick more explicitly/intentionally in a bunch of the other tests.

src/fusion_power_plant_tests.cc
dean-krueger: The point of this test was more to ensure that if users set different Li7 contribution the amount of tritium they get out doesn't change, which it shouldn't. Thinking about our design decision, I don't fully remember why we added the ability to change how much the Li-7 contributes to the breeding reaction, but I think it had to do with some nuance of blanket design? I remember Matthew running a whole openMC calculation which we based the default on, but beyond that it was so long ago that I forget.

gonuke: Ultimately, the blanket waste will have different compositions based on this, I guess.

gonuke (changes requested):

src/fusion_power_plant_tests.cc
```diff
@@ -37,18 +86,309 @@ TEST_F(FusionPowerPlantTest, Print) {
...
+
+
+  int simdur = 2;
+  cyclus::MockSim sim = InitializeSim(config, simdur);
+  
+  int id = sim.Run();
+
+  std::vector<Cond> conds;
+  conds.push_back(Cond("Time", "==", std::string("1")));
+  QueryResult qr = sim.db().Query("FPPInventories", &conds);
```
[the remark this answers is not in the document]
gonuke: I mean it to replace the three lines starting with the declaration of the `conds`

gonuke (approved):
This looks good now.  Thanks for all the improvements @dean-krueger

PR #30, opened 2025-01-30
Fixed tritium_storage never decaying

### LINKED ISSUE #27: Tricycle seems to be giving incorrect internal inventories
There seems to be some discrepancy between what tricycle is reporting and what we expect it to be giving us based on "theory". I don't have the plot on me, but I'll add it as a comment on this issue for more details. Either way, this needs to be fixed before anyone starts using tricycle.

dean-krueger:
Closes #27 

Screenshot 2025-01-30 at 10 09 55 AM

Some work was done to determine why Tricycle's internal inventories weren't matching what we expected them to be theoretically, and through that work it was discovered that the `tritium_storage` ResBuf was never decaying. Two lines of code were added to the very beginning of the `LoadCore()` function to fix this.

```
tritium_storage.Decay();
ExtractHelium();
```

This forces a decay of tritium_storage right before new tritium is bred, which avoids it always having its dt = 0 from always being "fresh" after the burn-breed cycle (this is what was causing the issue). It should be logic safe from the perspective of decay, since double decaying in the same time step does nothing (Decay() sets prev_decay_time to the current time making dt = 0 after decay). After this change, we have the following:

Inventory Behavior Comparison

Note:  this may cause a logic issue with how we check whether there's enough tritium in the system to operate in the edge case where a user sets their `reserve_inventory` too low (too close to the amount of tritium that actually gets burned in the system). I haven't thought much about how to deal with this, but figured it would be best to get this PR in and then maybe make a different issue later to fix that if we decide it's a problem. If the issue occurs it should crash the sim (`LoadCore()` will try to extract more tritium that exists from `tritium_storage` and I think that throws an error), instead of quietly giving incorrect results.

### DIFF
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -103,6 +103,9 @@ void FusionPowerPlant::EnterNotify() {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tick() {
 
+  DecayInventories();
+  ExtractHelium();
+
   if (ReadyToOperate()) {
     fuel_startup_policy.Stop();
     fuel_refill_policy.Start();
@@ -114,9 +117,6 @@ void FusionPowerPlant::Tick() {
     // Some way of leaving a record of what is going wrong is helpful info I think
     // Use the cyclus logger
   }
-  
-  DecayInventories();
-  ExtractHelium();
   
   double excess_tritium = std::max(tritium_storage.quantity() - 
                                   (reserve_inventory + SequesteredTritiumGap())
@@ -188,6 +188,7 @@ bool FusionPowerPlant::ReadyToOperate() {
   double required_storage_inventory = SequesteredTritiumGap();
   if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
     required_storage_inventory += reserve_inventory;
+    required_storage_inventory *= tritium_startup_fraction;
   } else {
     required_storage_inventory += fuel_usage_mass;
   }
@@ -209,10 +210,6 @@ void FusionPowerPlant::LoadCore() {
   if (SequesteredTritiumGap() > cyclus::eps_rsrc()) { 
     sequestered_tritium->Absorb(tritium_storage.Pop(SequesteredTritiumGap()));
   }
-
-  // Force decay of storage_inventory to avoid it resetting dt to 0
-  tritium_storage.Decay();
-  ExtractHelium();
 
   CycleBlanket();
   incore_fuel->Absorb(tritium_storage.Pop(fuel_usage_mass));
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -102,7 +102,6 @@ void FusionPowerPlant::EnterNotify() {
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::Tick() {
-  
 
   if (ReadyToOperate()) {
     fuel_startup_policy.Stop();
@@ -206,19 +205,17 @@ bool FusionPowerPlant::ReadyToOperate() {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::LoadCore() {
   
+  // Squash runs into issues when you give it zero, so we need to check frist
+  if (SequesteredTritiumGap() > cyclus::eps_rsrc()) { 
+    sequestered_tritium->Absorb(tritium_storage.Pop(SequesteredTritiumGap()));
+  }
+
   // Force decay of storage_inventory to avoid it resetting dt to 0
   tritium_storage.Decay();
   ExtractHelium();
 
   CycleBlanket();
-
-  // Squash runs into issues when you give it zero, so we need to check frist
-  if (SequesteredTritiumGap() > cyclus::eps_rsrc()) { 
-    sequestered_tritium->Absorb(tritium_storage.Pop(SequesteredTritiumGap()));
-  }
   incore_fuel->Absorb(tritium_storage.Pop(fuel_usage_mass));
-
-  
 
 }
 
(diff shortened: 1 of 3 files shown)


src/fusion_power_plant.cc
```diff
@@ -205,13 +205,18 @@ bool FusionPowerPlant::ReadyToOperate() {
...
   
-  CycleBlanket();
-
   // Squash runs into issues when you give it zero, so we need to check frist
   if (SequesteredTritiumGap() > cyclus::eps_rsrc()) { 
     sequestered_tritium->Absorb(tritium_storage.Pop(SequesteredTritiumGap()));
   }
+
+  // Force decay of storage_inventory to avoid it resetting dt to 0
+  tritium_storage.Decay();
```
gonuke, line 214: Why doesn't this happen before fixing the sequestered gap?

gonuke (changes requested):
What if we just moved `DecayInventories()` and `ExtractHelium()` to the beginning of the `Tick()`?

It might make the most sense for consistency, and it also matches the notion that the `Tick()` marks the beginning of a new time step - time has passed...

gonuke:
Also... if we always extract He immediately after decaying inventories, should we combine those two methods??? Not sure but worth discussing...

dean-krueger:
> What if we just moved `DecayInventories()` and `ExtractHelium()` to the beginning of the `Tick()`?
> 
> It might make the most sense for consistency, and it also matches the notion that the `Tick()` marks the beginning of a new time step - time has passed...

Doing this makes it so that we have to start predicting overbuy for ReadyToOperate() to function, which I tried that one time and ended up being bad for two reasons:

1. It was hard to do accurately (getting really close was easy, just multiply by lambda or whatever)
2. It changed the behavior in a way that was not explicitly defined by the user (they say buy 1kg, but *secretly* we buy 1.055kg because we know some of it is going to decay, or something like that). This seems like a dangerous game to play in modeling.

> Also... if we always extract He immediately after decaying inventories, should we combine those two methods??? Not sure but worth discussing...

I see two conflicting philosophies here:

1. The one you're thinking about: we always extract helium after calling DecayInventories, so maybe DecayInventories calls extract helium
2. The other one: decaying and extracting helium are fundamentally different operations wrt fusion energy system design (so far as I understand), and so explicitly doing them separately potentially makes any future, higher fidelity, functionality potentially easier to implement/more readable.

Agree that this is probably worth a conversation, but those are my ¢2

src/fusion_power_plant.cc
dean-krueger: This is a good question! My reasoning for this, after initially decaying *before* sequestration like you're suggesting, is basically "because it works when you do it this way".

The long answer is: if you decay the tritium before sequestering it, it goes into sequestered_tritium "decayed" and then the tritium in sequestered tritium doesn't decay again later and there's a double entry of sequestered_equilibrium which messes up the unit tests (and doesn't make sense). Doing it in this order avoids that, and keeps the sequestered_tritium operating as it did before, while still adding the new decay behavior that fixes overall inventory. Hopefully that sort of makes sense? I'd be happy to explore this more during one of the fuel-cycle meetings. There is a certain level to which I don't fully know why it works this way, I just know what happens because I got it to work so I stopped investigating, but if it's potentially problematic I can certainly dive back in.

gonuke:
Plan developed during discussion:
* Decay all tritium inventories at the beginning of a Tick for consistency - all T inventories will be accurate for that point in simulated time
* This introduces a challenge for startup because we will always be one time step's decay short of enough material
* Change the startup condition to be lower than `reserve_inventory + sequestered_equilibrium`
* TBD: does the user provide a different `startup_reserve_inventory` or some fraction < 1 to indicate how little T is necessary to startup?

dean-krueger:
Just to confirm visually that this new method is performing the same as the old one:

decay_compoarison

src/fusion_power_plant.h
```diff
@@ -106,6 +106,16 @@ class FusionPowerPlant : public cyclus::Facility  {
   }
   double sequestered_equilibrium; 
 
+  #pragma cyclus var { \
+    "default": 0.9, \
+    "doc": "Fraction of desired startup tritium inventory required. ", \
```
gonuke, line 111:
```suggestion
    "doc": "Fraction of desired tritium reserve inventory required for initial startup. ", \
```

src/fusion_power_plant.h
```diff
@@ -106,6 +106,16 @@ class FusionPowerPlant : public cyclus::Facility  {
...
+    "tooltip": "Gives flexibility accounting for decay", \
+    "units": "Dimensionless", \
```
gonuke, line 113: Not sure if this matters?
```suggestion
    "units": "dimensionless", \
```

src/fusion_power_plant.h
```diff
@@ -106,6 +106,16 @@ class FusionPowerPlant : public cyclus::Facility  {
   }
   double sequestered_equilibrium; 
 
+  #pragma cyclus var { \
+    "default": 0.9, \
+    "doc": "Fraction of desired startup tritium inventory required. ", \
+    "tooltip": "Gives flexibility accounting for decay", \
```
gonuke, line 112:
```suggestion
    "tooltip": "Fraction of reserve required for startup", \
```

gonuke (changes requested):
This all looks good now - I left a few suggestions to make the documentation more clear (to me at least)

src/fusion_power_plant.h
dean-krueger: The actual implementation multiplies the entire startup (sequestered_equilibrium + reserve) by the fraction, not just the reserve, which is why I said "desired startup tritium" instead of "... reserve ...". Let me know if that needs to change.

gonuke: Yeah... but the user never actually declares a "startup" inventory in the input file.  Maybe this should be:
```suggestion
    "doc": "Fraction of desired startup ( = reserve + sequestered) tritium inventory required for initial startup. ", \
```

It also reads oddly because why would you not need 100% of the startup inventory for startup?  If you don't actually need that much for startup, why call it the startup inventory.... maybe its just a vocabulary thing?

dean-krueger: That works, I'll commit! Just wanted to make sure we were being clear.

gonuke:
There are some merge conflicts apparently? Otherwise ready to merge...

dean-krueger:
Merge conflicts (should be) taken care of!

gonuke (approved):

PR #31, opened 2025-01-30
Code Formatting PR

### LINKED ISSUE #29: Tricycle needs to be formatted correctly
There is currently no formatting done on the Tricycle source code. We should add a clang-format file and make Tricycle use it (except on the header file, because currently that breaks things)

dean-krueger:
The code was, previously, not formatted with clang-format. This PR fixes that. Ran unit tests and a sample problem to make sure nothing fundamentally broke.

Closes #29 

Note, it was not possible to format the .h file, since it would break the #pragma lines.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,11 @@
+---
+Language:        Cpp
+BasedOnStyle:  Google
+# only allow single line functions if they are inlined in class def'n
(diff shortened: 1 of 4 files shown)


gonuke (approved):
Thanks @dean-krueger  This looks ok and passes tests.  

Can you please make a separate draft PR to show what it does to the header file?

PR #32, opened 2025-02-04
formatted all code in src, including header file

dean-krueger:
Doing this, specifically formatting the header file with clang-format, gives the following error:

```
SyntaxError: For the statement:


#pragma cyclus var{                                                          "doc" : "Achievable system tritium breeding ratio before decay",         "tooltip" : "Achievable system tritium breeding ratio before decay",     "units" : "non-dimensional", "uitype" : "range",                         "range" : [0, 1e299], "uilabel" : "Tritium Breeding Ratio" }
This appears to be a cyclus pragma but has an incorrect form!
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,11 @@
+---
+Language:        Cpp
+BasedOnStyle:  Google
+# only allow single line functions if they are inlined in class def'n
+AllowShortFunctionsOnASingleLine: Inline
+# do not force new line after template declarations
+AlwaysBreakTemplateDeclarations: false
+# Automatically detect whether a given fuction's arguments are one-per-line
+ExperimentalAutoDetectBinPacking: true
+# Use C++11 standard
+Standard:        Cpp11
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -1,13 +1,13 @@
 #include "fusion_power_plant.h"
 
-using cyclus::Material;
-using cyclus::Composition;
 using cyclus::CompMap;
-using cyclus::IntDistribution;
+using cyclus::Composition;
 using cyclus::DoubleDistribution;
-using cyclus::FixedIntDist;
 using cyclus::FixedDoubleDist;
+using cyclus::FixedIntDist;
+using cyclus::IntDistribution;
 using cyclus::KeyError;
+using cyclus::Material;
 using cyclus::toolkit::ResBuf;
 
 namespace tricycle {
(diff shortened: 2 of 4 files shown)


gonuke:
Do you get any information about which line number causes this particular error? Or one example of this error?

gonuke:
I think its because there is no space after `var` which feels like something we can train clang to get right

dean-krueger:
Yep, it's definitely the space after var. I went and checked the cycpp file and the regex used to flag "wrong" pragmas is: 

```
\s*#\s*pragma\s+cyclus(.*)
```

where the `\s+` after `pragma` means "look for one or more spaces". If this was changed to `\s*` it would look for any number of whitespace characters (I think including zero), but I think we'd have to be careful about making sure to change this in all the places it needs to be changed in to avoid simply not flagging errors when they occur. This is going to take some care and thought, but is something I can futz around with today.

dean-krueger:
Update: it's definitely the space after `var`. cycpp.py specifies that the format of these needs to be `#pragma cyclus <decorator> [args]`, and in the case of the `var` decorator, it expects `#pragma cyclus var {dict}`. I'm guessing this is because somewhere else a string split happens on ' ' and so the error message I was seeing was flagging the lack of space as a problem. Specifically, it uses this regex

```
regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)")
```

Where `\s+var\s+` means that there needs to be exactly one space before and after the `var` decorator. If this was changed to `\s*var\s*` it would be able to take any amount of whitespace around var (including zero I think), but that would probably mess with the hypothetical string split I think (but have not confirmed) is going on somewhere else to actually parse these pragmas.

dean-krueger:
Another Update: It seems like clang-format doesn't even do the same thing to all the `#pragma` lines? Sometimes it chooses to remove the space in `var {` and sometimes it doesn't. It also doesn't fully shorten all the lines within the var dicts, which is odd. Specifically, after running clang-format I have:

```
#pragma cyclus var{"doc" : "Nameplate fusion power of the reactor",      \
                   "tooltip" : "Nameplate fusion power", "units" : "MW", \
                   "uitype" : "range",                                   \
                   "range" : [0, 1e299], "uilabel" : "Fusion Power" }
  double fusion_power;
```
AND
```
#pragma cyclus var { \
    "doc": "Minimum tritium inventory to hold in reserve in case of tritium recovery system failure", \
    "tooltip": "Minimum tritium inventory to hold in reserve", \
    "units": "kg", \
    "uilabel": "Reserve Inventory" \
  }
  double reserve_inventory;
```

Not certain about what's going on here.

gonuke:
This particular error is actually generated on line 411 of cycpp, driven by the split on line 410, but the regex you identified would ultimately also be relevant. We could probably come up with a way to make these regex compliant, but  I don't know what other places we need to consider, though.

PR #41, opened 2025-08-12
Tricycle development branch

trysarahtops:
This includes post-processing scripts, Pearson and Kovari CANDU histories, and CANDU code snippet.

### DIFF
--- a/100MW_FPPwGlobalSupply.xml
+++ b/100MW_FPPwGlobalSupply.xml
@@ -1,0 +1,784 @@
+  <simulation>
+  <control>
+    <duration>2000</duration>
+    <startmonth>1</startmonth>
+    <startyear>2006</startyear>
+    <decay>lazy</decay>
+    <explicit_inventory>true</explicit_inventory>
+  </control>
+
+  <archetypes>
+    <spec><lib>cycamore</lib><name>DeployInst</name></spec>
+    <spec><lib>cycamore</lib><name>Source</name></spec>
+    <spec><lib>cycamore</lib><name>Storage</name></spec>
+    <spec><lib>cycamore</lib><name>Sink</name></spec>
+    <spec><lib>agents</lib><name>NullInst</name></spec>
+    <spec><lib>agents</lib><name>NullRegion</name></spec>
+    <spec><lib>tricycle</lib><name>FusionPowerPlant</name></spec>
+  </archetypes>
+
+  <!--Start of the Commodities Section-->
+  <commodity>
+    <name>Tritium</name>
+    <solution_priority>1.0</solution_priority>
+  </commodity>
+  <commodity>
+    <name>Helium_3</name>
+    <solution_priority>1.0</solution_priority>
+  </commodity>
+
+
+  <!--Start of the Recipe Section-->
+  <recipe>
+    <name>Tritium</name>
+    <basis>atom</basis>
+    <nuclide>
+        <id>10030000</id>
+        <comp>1</comp>
+    </nuclide>
+  </recipe>
+
+  <recipe>
+    <name>Helium_3</name>
+    <basis>atom</basis>
+    <nuclide>
+        <id>20030000</id>
+        <comp>1</comp>
+    </nuclide>
+  </recipe>
+
+  <recipe>
+    <name>Lithium</name>
+    <basis>atom</basis>
+    <nuclide>
+      <id>30060000</id>
+      <comp>0.3</comp>
+    </nuclide>
+    <nuclide>
+      <id>30070000</id>
+      <comp>0.7</comp>
+    </nuclide>
+  </recipe>
+
+  <facility>
+    <name>OnePlant</name>
+    <config>
+      <FusionPowerPlant>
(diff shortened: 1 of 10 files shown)


100MW_FPPwGlobalSupply.xml
```diff
@@ -0,0 +1,784 @@
...
+  <facility>
+    <name>Bruce A1 (R)</name>
+    <config>
+        <Source>
+            <outcommod>Tritium</outcommod>
+            <outrecipe>Tritium</outrecipe>
+            <throughput>0.016467</throughput>
+        </Source>
+    </config>
+  </facility>
```
gonuke, line 236: We should discuss whether it is valuable enough to have each of these named separately (maybe for debugging), or whether we can define generic CANDUs at the appropriate throughputs and then just start them in the appropriate year.

KovariCanadianCANDUs.xml
gonuke, line 1: It would be nice to also include a `Region` and `DeployInst` for all of these that can also be imported, perhaps to a separate part of the file.  However, that makes me realize that the timing of deployment in a `DeployInst` assumes a start year/month which limits how easy it is to include deployment information with XInclude. cyclus/cycamore#664

SourcetoPlanttoSink_v1.xml
```diff
@@ -0,0 +1,150 @@
...
+  <facility>
+    <name>One Storage</name>
+    <config>
+      <Storage>
+        <in_commods>
+            <val>Tritium</val>
+        </in_commods>
+        <out_commods>
+            <val>Tritium</val>
+        </out_commods>
```
gonuke, line 99: Isn't this prone to self-trading?

SourcetoSeperationstoStoragetoPlant_v1.xml
```diff
@@ -0,0 +1,176 @@
...
+ <facility>
+  <name>Separations Facility</name>
+  <config>
+    <Separations>
+      <feed_commods>
+        <val>Tritium</val>
+      </feed_commods>
+      <feed_commod_prefs>
+        <val>1.0</val>
+        <val>1.0</val>
```
gonuke, line 115: I think this needs to have the same number of entries as `<feed_commods>`

SourcetoSeperationstoStoragetoPlant_v1.xml
```diff
@@ -0,0 +1,176 @@
...
+      </feed_commod_prefs>
+      <feedbuf_size>8.121</feedbuf_size>
+      <throughput>1e+3</throughput>
+      <leftover_commod>Helium_3</leftover_commod>
+      <leftoverbuf_size>1e+3</leftoverbuf_size>
+      <streams>
+        <item>
+          <commod>Separated_Material</commod>
```
gonuke, line 123: I think giving these more meaningful names might help interpretation.
```suggestion
          <commod>Separated_Helium3</commod>
```

SourcetoSeperationstoStoragetoPlant_v1.xml
```diff
@@ -0,0 +1,176 @@
...
+          <info>
+            <buf_size>5e+3</buf_size>
+            <efficiencies>
+              <item>
+                <comp>2003000</comp>
+                <eff>0.99</eff>
```
gonuke, line 129: How does Tricycle deal with potentially getting feed that has a little He3 in it?

gonuke (changes requested):
Thanks for this PR @trysarahtops !! Seeing this all in with my review hat on really gets me thinking about ways to make Cyclus easier to use (which are some of my comments). 

Some things we should discuss generically:
* where should these sample/benchmark (?) files live, i.e. which directories?
* can we use the CANDU xml snippets in our input files rather than duplicating them there?
* an we include the deployment histories and archetype specifications with the CANDU xml snippets for completeness?
* a couple of Cyclus & Cycamore improvements that occur to me

It is also apparent to me that we need to start with smaller PRs so that we can tackle issues in a single input file, and then let you apply those to subsequent files.

PearsonCANDUs.xml
```diff
@@ -0,0 +1,758 @@
...
+    </config>
+  </facility>
+
+    <facility>
+    <name>Bruce A2</name>
+    <config>
+        <Source>
+            <outcommod>Tritium</outcommod>
+            <outrecipe>T</outrecipe>
+            <throughput>0.010833</throughput>
```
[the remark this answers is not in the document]
trysarahtops: Pearson uses an standard value average on all data based on theoretical reactor physics while Kovari uses calculations from the actual reactors.

gonuke:
Let's put these files in a structure like:
```
examples/
    - includes/
        - CANDU XML snippets, one per country with fully defined region and institution(s)
    - PearsonBenchmark/
        - full input file, importing CANDU XML snippets as approriate
        - some post-processing script and/or Python notebook
    - KovariBenchmark
        - full input file, importing CANDU XML snippets as approriate
        - some post-processing script and/or Python notebook
    - samples/
        - various simpler demonstrations which may or may not include CANDUs

PR #59, opened 2025-10-22
Digitized global inventory

trysarahtops:
Digitized csv data for Kovari Scenario A, Pearson Supply 1 Demand A, Abdou with ITER, and Abdou w/o fusion


gonuke:
Thanks for contributing this @trysarahtops.  Can you make a plot (or 2 or 3 if that makes more sense) with these data and paste it in the discussion of this PR?

Also, I think we should place these in a subdirectory somewhere to keep things organized. Maybe something like `reference_data` unless you have a better suggestion.

trysarahtops:
DigitizedTritiumData

This plot shows all of the digitized data files plotted together.

PR #70, opened 2026-01-08
Global input files

trysarahtops:
Update the assumptions and files used to recreate the pessimistic model of tritium inventory generated by Kovari et. al using Tricycle.

Here is an image comparing the model to the digitized data.

KovariComparison


gonuke (approved):
This looks great.  I'm ready to merge this one and the others are close.

PR #71, opened 2026-02-06
Template files for python script to generate FPP

trysarahtops:
Added templates and test files to generate xml code snippets for scenario integration. I attempted to use placeholder brackes {} instead of the xml ElementTree.


Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,37 @@
+import csv
+
+# enter the csv or text file of FPP data
+
+FPP = input("Enter FPP file name: ")
+Deploy = input("Enter deployment timeline file name: ")
```
gonuke: Let's add `argparse` to make these command-line variables instead.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,37 @@
...
+    Reads a CSV file into a python list
+    """
+
+    data = []
+
+    with open(filename, mode='r', newline='', encoding='utf-8') as file:
+        reader = csv.reader(file)
+        for row in reader:
+            data.append(row)
+    return data[1] # Returns the values of FPP input data
```
gonuke: This only returns the first entry - maybe that's just for testing right now?  I also note that your test files only have one row.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,37 @@
...
+
+csv_list = read_csv_to_list(FPP)
+
+f = open('FPPTemplate.txt', 'r')
+plant = f.read()
+formatted = plant.format(*csv_list)
```
gonuke: If we put this in a loop for each row, we could just append the additional facilities

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,37 @@
...
+
+with open("FPP.xml", "w") as c:
+    c.write(formatted)
+
+timeline = read_csv_to_list(Deploy)
+
+t = open('Deployment.txt', 'r')
+time = t.read()
+comm = time.format(*timeline)
```
gonuke: This doesn't scale as well for cases when we have many deployments.  We'll want to discuss the best plan here.

gonuke (changes requested):
This is a good start, and as a first "blank slate" scripting project a great place to focus on some best practice.

In general, I'll recommend you take a look at [this repo](https://github.com/cnerg/py-template).  The REAMDE discusses some best practices and there is [a specific files](https://github.com/cnerg/py-template/blob/main/acme_corp/do_task.py) that demonstrates some of those practices.

Some things that jump out at me here:
* using the `argparse` module to handle user interaction
* creating a set of small functions with specific purposes that are called from a main routine
   * ideally, very little code is called at the module scope and all of it should be in functions
* use the `if __name___ == "__main__":` pattern at the end of the module
* as the above develops there may be more descriptive variable names that make sense


Beyond general best practices - another tip:
* the CSV DictReader can read your CSV files into dictionaries so that you can refer to entries by their header name rather than just by position.  This might make the template formatting more robust because you can put a specific item in a specific position without worrying about the order getting messed up.
* We may want to assemble lists of prototypes, start_years, lifetimes, and builds to substitute in the deployment

Scenarios/GenFPPscript.py
trysarahtops: I took a look at the repo referenced above and attempted to implement some of the suggested changes. I am still looking through the repo and README and will make additional edits/updates as necessary.

gonuke: It's already looking so much better!  I'll hold off on commenting until you request another review, but feel free to reach out to me or others with questions.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,47 @@
...
+# function to turn csv file into list
+
+def read_csv_to_list(filename):
+    """
+    Reads a CSV file into a python list
+    """
+
+    data = []
+
+    with open(filename, 'r') as fp:
```
MicahGale: A few thoughts here.

1. Knowing the line numbers ahead of time isn't necessary. If you do need the line number you can do:

``` python
with open() as file:
    for line, line_number in enumerate(file):
         ...
```

2. It's best to make the reader object once.
 
``` python
with open() as file:
    reader = csv.reader(file)
    for row, line_number in enumerate(reader):
         ...
```

3. You are discarding the header row (for obvious reasons), but [`csv.dictreader`](https://docs.python.org/3/library/csv.html#csv.DictReader) can handle it make your life so much easier. With this code you will get a list of dictionaries. So each dictionary in the list represents a single fusion power plant

``` python
with open() as file:
    reader = csv.DictReader(file)
    rows = []
    for row, line_number in enumerate(reader):
         rows.append(row)
         # example 
         print(row["name"], row["power"])
```

Finally: I'm lazy and just have pandas do everything for me with [`pd.read_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html).

This will generate a pandas dataframe with named columns matching the header row. The amount of options ... is insane. So let me know if you have any questions.

``` python
    import pandas as pd
    return pd.read_csv(filename)
```

Without more context of an example spreadsheet I can't give too much more arguments for this function.

Scenarios/FPPTemplate.txt
```diff
@@ -0,0 +1,24 @@
+  <facility>
+    <name>{}</name>
```
MicahGale: I would recommend naming all your [format specifications](https://www.w3schools.com/python/ref_string_format.asp) like:

``` suggestion
    <name>{name}</name>
```

The format specification is <name>:<format>. So say you wanted to round of a variable, `foo` as a number to two decimal places you would say: `{foo:.2f}`.

To pass these arguments I would suggest making sure you use the same column header names from the csv. That way you can use the dictionary returned from the csv reader.

``` python
fpp_plant_dict # came from magical place previously
file_string.format(**fpp_plant_dict)
```

This uses some syntax sugar where a dictionary is expanded to be a bunch of keyword arguments to the function.

MicahGale (commented):
@trysarahtops I hope these suggestions help. Let me know if you have any questions.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,47 @@
...
+with open("FPP_spec.xml", "w") as fplant:
+     for i in range(len(fpp_list)):
+         fppspec = open('FPPTemplate.txt', 'r').read().format(*fpp_list[i])
+         fplant.write(fppspec)
+         
+# fill the Deployment template with regions and commisioning timeline
+with open("FPP_dep.xml", "w") as comm:
+   for i in range(len(dep_list)):
+    dep_time = open('Deployment.txt', 'r').read().format(*dep_list[i])
+    comm.write(dep_time)
```
gonuke: These should either be part of the `if __name__` block or be in functions that are called from that block (the latter is better)

gonuke: Also, since each of these `with` blocks had identical code, but different variables, there is probably a function that does it once, that can be called twice with the different data.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,47 @@
...
+
+import csv
+import argparse
+
+#Enter FPP spec and Deployment commissioning input files
+
+parser = argparse.ArgumentParser(description='Generate FPP deployment scenario for tricycle')
+parser.add_argument('-f','--fpp', type=str, help='FPP spec file')
+parser.add_argument('-d','--dep', type=str, help='FPP deployment file')
+args = parser.parse_args()
```
gonuke: Put this in a function that is called from the `if __name__` block

gonuke (changes requested):

Scenarios/GenFPPscript.py
MicahGale: Yes you can open multiple files as once in a `with` block:

``` python
with open("FPP_spec.xml", "w") as fplant,  open("FPP_dep.xml", "w") as comm:
    ...
```

Scenarios/DeployIn.csv
```diff
@@ -0,0 +1,5 @@
+region name, institution, prototypes, build_times, lifetimes, n_build
```
gonuke: Need to make sure the labels in this header match the labels used in the template files

Scenarios/Deployment.txt
gonuke, line 1: I think this file has been replaces by the separate institution and region files

Scenarios/FPPInput.csv
```diff
@@ -0,0 +1,2 @@
+name,fusion_power,TBR,Reserve Inventory,Sequestered Equilibrium,Fuel Incommodity,Lithium 7 Contribution,Refuel Mode,Buy Quantity,Buy Frequency,Helium 3 Outcommodity,Blanket inrecipe,Blanket Incommodity,Blanket Outcommodity,Blanket Size,Blanket Turnover Fraction,Blanket Turnover Frequency
```
gonuke: need to make sure these labels match the values used in the template file

Scenarios/FPPTemplate.txt
```diff
@@ -0,0 +1,24 @@
+  <facility>
+    <name>{name}</name>
+    <config>
+      <FusionPowerPlant>
+        <fusion_power>{fusion_power}</fusion_power>
+        <TBR>{TBR}</TBR>
+        <reserve_inventory>{reserve_inventory}</reserve_inventory>
+        <sequestered_equilibrium>{}</sequestered_equilibrium>
```
gonuke: need to update all of these empty template variables with labeled template variables that match the CSV headers

Scenarios/FPPTemplate.txt
gonuke, line 1: minor thought: we could rename the template files to end with ".tmpl" as their extension to make the purpose even more clear

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,128 @@
...
+        result.append(spec)
+    return '\n\n'.join(result)  
+
+def build_XML_snippet():
+    """
+    Main function to build the FPP and Deployment XML snippets. Reads in the FPP
+    spec and deployment commissioning timeline from csv files, fills in the
+    template files, and writes the resulting XML snippets to output files.                         
+    """
+    add_parse()
```
gonuke:
```suggestion
    args = add_parse()
```

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,128 @@
...
+    parser = argparse.ArgumentParser(description='Generate FPP deployment scenario for tricycle')
+    parser.add_argument('-f','--fpp', type=str, help='FPP spec file')
+    parser.add_argument('-d','--dep', type=str, help='FPP deployment file')
+    return parser.parse_args()
+
+# function to turn csv file into list
+
+def read_csv_to_list(filename):
+    """
+    Reads a CSV file into a python list
```
gonuke:
```suggestion
    Reads a CSV file into a python list of dictionaries, with dictionaries keys based on the CSV header values
```

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,128 @@
...
+    
+
+    xml_string = template.format(inst_name=inst_name, prototypes=prototypes, 
+                                 build_times=build_times, lifetimes=lifetimes, 
+                                 n_build=n_build)
+
+    return xml_string
+
+
+def fill_template(data_list, template):
```
gonuke: I am not sure this method has a lot of value.  It probably makes more sense to assemble a list of XML snippets directly in `build_XML_snippet()` and then process them with the join as you write them.  (see comment below)

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,128 @@
...
+    spec and deployment commissioning timeline from csv files, fills in the
+    template files, and writes the resulting XML snippets to output files.                         
+    """
+    add_parse()
+
+    fpp_list = read_csv_to_list(args.fpp)
+    # fill the FPP template file with specs
+    with open('FPPTemplate.txt', 'r') as template_file:
+        template = template_file.read()
+    fpp_string = fill_template( fpp_list, template)
```
gonuke:
```suggestion
    fpp_xml_list = []
    for fpp in fpp_list:
        fpp_xml_list.append(template.format(*fpp))
    fpp_string = "\n\n".join(fpp_xml_list)
```

gonuke (changes requested):
Here some thoughts to follow up from the work we did earlier this week.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Reads a CSV file into a python list of dictionaries with dictionary keys 
+    based on the CSV header values
+    """
+
+    data = []
+
+    with open(filename, mode='r', newline='', encoding='utf-8') as file:
+        reader = csv.DictReader(file)
+        #header = next(reader) # Reads and discards the first row
```
gonuke: We can delete these lines we don't need anymore

trysarahtops:
The script now generates output deployment and FPP xml files in conjunction with necessary template and input files. There is still a small issue with formatting that I am working out for the deployment output.

Scenarios/GenFPPscript.py
```diff
@@ -0,0 +1,122 @@
...
+
+def process_deployment_data(data_list):
+
+    deployment_map = {}
+
+    for entry in data_list:
+        region = entry['region_name']
+        if region not in deployment_map:
+            deployment_map[region] = {}
+            institution = entry['institution']
```
gonuke: This line should not be indented, it should not be nested in the conditional `if` block

gonuke (changes requested):
A couple of little changes

gonuke:
It seems like the headers in the FPP CSV file need to be udpated to match their template, still?

gonuke (approved):
This LGTM - thanks @trysarahtops

PR #73, opened 2026-03-14
Create GTSwITER.xml

trysarahtops:
Input file used to model the tritium startup in Pearson. The launch year is considered 2035 and runs for 12 years.

The image below shows the changes in tritium inventory with the startup of ITER.

Pearson Tritium Inventory with ITER

This image (below) shows the excess tritium produced in ITER with a TBR of 1.05.
Pearson TBR 105

The image below shows the excess tritium produced in ITER with a TBR of 1.15.
Pearson TBR 115


gonuke:
ITER is a net consumer of tritium.  I expect the TBR to be 0.  Do we have references that suggest otherwise?

Scenarios/GTSwITER.xml
```diff
@@ -0,0 +1,274 @@
...
+      <Source>
+        <outcommod>Li7</outcommod>
+        <outrecipe>Lithium</outrecipe>
+        <throughput>500</throughput>
+      </Source>
+    </config>
+  </facility>
+
+  <facility>
+    <name>ITER</name>
```
gonuke: Do we have a reference for these facility characteristics?  You have called it ITER, but I don't expect ITER to ever operate like a power plant.  Maybe this is referring to a different 500 MWe fusion power plant?

gonuke (changes requested):
Thanks @trysarahtops - I'm just a little curious about the goal of the new facilities.  Perhaps a README (ro additions to the README) will help explain the goal of this scenario.

Scenarios/GTSwITER.xml
trysarahtops: I used the FPP to sample the tritium breeding expected in ITER not necessarily to operate like a fusion power plant. When I used the timeline set out by Pearson, it took into account the tritium used from the global supply, but not the tritium produced in ITER. Maybe I should approach this differently.

gonuke: ITER will not be a net breeder.  It has no breeding blanket and only a tiny fraction of the wall area will have test breeding modules.  ITER will be a pure sink, which is how you include it for the most part.

PR #74, opened 2026-03-31
Abdou with ITER model

trysarahtops:
I have included the files that were used to recreate the Abdou plot with ITER. The files include a markdown file, Canadian and Korean CANDU files, and ITER file, and a global inventory file. The previous ITER file was deleted.

Here is the data plotted along with the digitized plot.
image


Scenarios/Abdou/AbdouGlobalT.xml
```diff
@@ -0,0 +1,106 @@
...
+      </config>
+  </facility>
+
+  <facility>
+    <name>Start Supply</name>
+    <config>
+      <Source>
+        <outcommod>Tritium</outcommod>
+        <outrecipe>T</outrecipe>
+        <throughput>18.5</throughput>
```
gonuke: I think we should specify `inventory_size` as well (instead?) for robustness

Scenarios/Abdou/AbdouGlobalT.xml
gonuke, line 1: We should put the facilities and region from this global supply in an XML snippet, too, to make it easier to include in multiple input files.  Since this info is repeated in the 2 files you have here, it's a good practice to avoid them accidentally diverging.

gonuke (changes requested):
This looks pretty good.  A couple of quick suggestions should cleat this up, but we should also reconcile with some older PRs.

Scenarios/Abdou/AbdouGlobalT.xml
trysarahtops: To clarify, would that look like this?

```
<facility>
    <name>Start Supply</name>
    <config>
      <Source>
        <outcommod>Tritium</outcommod>
        <outrecipe>T</outrecipe>
        <throughput>18.5</throughput>
        <inventory_size>18.5</inventory_size>
      </Source>
    </config>
  </facility>
```

Since this facility only exists for one time step and is a source, the inventory_size and throughput should match, as I understand it.

gonuke (approved):
This looks great - thanks @trysarahtops

PR #75, opened 2026-04-24
Decay storage

trysarahtops:
The decay storage facility has been updated to include the max tritium inventory to be an optional user input and throughput has been added as an optional user input; both take `CY_LARGE_DOUBLE` as the default value. The unit tests have been updated to reflect these changes.

The Pystart/stop and tritium recipe was removed from the unit testing.


src/decay_storage.cc
```diff
@@ -0,0 +1,77 @@
...
+  fuel_tracker.Init({&tritium_storage}, cyclus::CY_LARGE_DOUBLE);
+
+  bool is_bulk = true;
+
+  tritium_storage = cyclus::toolkit::ResBuf<cyclus::Material>(is_bulk);
+  helium_storage = cyclus::toolkit::ResBuf<cyclus::Material>(is_bulk);
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+std::string DecayStorage::str() {
```
gonuke: I don't think we need this method unless we're going to develop a custom string that represents a facility of this archetype (which we may want to do)

src/decay_storage.cc
```diff
@@ -0,0 +1,77 @@
...
+      ->AddVal("AgentId", id())
+      ->AddVal("Time", context()->time())
+      ->AddVal("TritiumStorage", tritium)
+      ->AddVal("HeliumStorage", helium)
+
+      ->Record();
+}
+
+void DecayStorage::ExtractHelium(
+    cyclus::toolkit::ResBuf<cyclus::Material>& inventory) {
```
gonuke: Since there is only one valid `ResBuf` from which we can extract helium, I think it's best if this takes no arguments

src/decay_storage.h
```diff
@@ -0,0 +1,133 @@
...
+/// @section detailed Detailed Behavior
+/// Each time step, the facility:
+/// - Tick: Decays all tritium inventory, then extracts accumulated helium-3
+///         (helium-3 is treated as a byproduct)
+/// - Tock: Records current inventories
+/// The tritium_storage buffer uses bulk storage mode for automatic material
+/// combining, and helium-3 is continuously separated and stored independently
+/// as a byproduct (not currently offered to market).
+class DecayStorage : public cyclus::Facility {
+ public:
```
gonuke, line 42: This class is too public.  We need to move some/much of this to `protected` and then possibly add `DecayStorageTest` as a friend class.  See `cycamore/storage` as an example.

gonuke (changes requested):
There are a few high-level design issues to discuss here, primarily that many of the member data/functions should be declared `private` or `protected`

src/decay_storage.cc
trysarahtops: I have removed this.

src/decay_storage.cc
trysarahtops: I am currently working on this.

src/decay_storage.h
trysarahtops: I have implemented this change.

trysarahtops:
There are a few more things that I would like to try before resubmitting.

trysarahtops:
I have updated the DecayStorage header file to remove any relics from the previous versions of the facility. However, I still was not able to get the DecayStorageTest header file to compile so I included all necessary items in the DecayStorageTest file.

src/decay_storage.cc
```diff
@@ -0,0 +1,71 @@
+// decay_storage.cc
+
+#include "decay_storage.h"
+
+#pragma cyclus exec from cyclus.system import CY_LARGE_DOUBLE, CY_LARGE_INT, CY_NEAR_ZERO
```
gonuke: I think this is superfluous - let's confirm

src/decay_storage.cc
```diff
@@ -0,0 +1,71 @@
...
+
+namespace tricycle {
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+DecayStorage::DecayStorage(cyclus::Context* ctx) : cyclus::Facility(ctx) {
+  // Required by DRE policies
+  fuel_tracker.Init({&tritium_storage, &helium_storage}, cyclus::CY_LARGE_DOUBLE);
```
gonuke: Probably only want to track tritium in this tracker.

src/decay_storage.cc
```diff
@@ -0,0 +1,71 @@
...
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+void DecayStorage::EnterNotify() {
+  cyclus::Facility::EnterNotify(); // call base function first
+  fuel_tracker.set_capacity(max_tritium_inventory);
+  buy_policy.Init(this, &tritium_storage, std::string("input"), &fuel_tracker).Set(incommod).Start();
+  sell_policy.Init(this, &tritium_storage, std::string("output")).Set(outcommod).Start();
+}
+
+void DecayStorage::RecordInventories(double tritium, double helium) {
```
gonuke: Instead of passing in the values of the `tritium` and `helium` quantities, we should just read these out of their `ResBuf`s.

src/decay_storage.cc
```diff
@@ -0,0 +1,71 @@
...
+  context()
+      ->NewDatum("StorageInventories")
+      ->AddVal("AgentId", id())
+      ->AddVal("Time", context()->time())
+      ->AddVal("TritiumStorage", tritium)
+      ->AddVal("HeliumStorage", helium)
```
gonuke:
```suggestion
      ->AddVal("TritiumStorage", tritium_storage.quantity())
      ->AddVal("HeliumStorage", helium_storage.quantity())
```

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, DatabaseRecording) {
+    // Test that inventories are properly recorded to the database
+  
+    std::string config = common_config;
+  
+    int simdur = 2;
+    cyclus::MockSim sim = InitializeSim(config, simdur);
```
gonuke: Let's skip the extra variable and just use the `common_config` directly..
```suggestion
    cyclus::MockSim sim = InitializeSim(common_config, simdur);
```

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, BasicMaterialFlow) {
+  // Test basic material flow: receiving tritium, storing it, and recording
+
+  std::string config = common_config;
+
+  int simdur = 2;
+  cyclus::MockSim sim = InitializeSim(config, simdur);
```
gonuke: ibid
```suggestion
  cyclus::MockSim sim = InitializeSim(common_config, simdur);
```

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, EmptyStorageBehavior) {
+  // Test behavior when storage is empty
+
+  std::string config = common_config;
+
+  int simdur = 2;
+  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:DecayStorage"), config,
```
gonuke: ibid

```suggestion
  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:DecayStorage"), common_config,
```

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, MaterialWithDecay) {
+  // Test with material that already contains some decay products
+
+  std::string config = common_config;
+
+  int simdur = 2;
+  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:DecayStorage"), config,
```
gonuke: ibid

```suggestion
  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:DecayStorage"), common_config,
```

gonuke (changes requested):
A few small changes - still need to think about the test with decay

gonuke:
Thanks for adding the `fuel_outcommod` to the FPP. In all of our discussion, we were assuming it was there, without realizing it wasn't!!

Can you make a separate PR that tackles just that aspect - adding the `fuel_outcommod` and associated tests - so that we can review it separately from the `DecayStorage` improvements?

PR #77, opened 2026-05-18
Updated decay storage files

trysarahtops:
Replaces #75

### DIFF
--- a/src/decay_storage_tests.cc
+++ b/src/decay_storage_tests.cc
@@ -1,0 +1,277 @@
+#include <gtest/gtest.h>
+
+#include <cmath>
+
+
+#include <string>
+#include "context.h"
+#include "facility_tests.h"
+#include "agent_tests.h"
+#include "decay_storage.h"
+
+using cyclus::CompMap;
+using cyclus::Cond;
+using cyclus::Material;
+using cyclus::QueryResult;
+using cyclus::toolkit::MatQuery;
+using tricycle::DecayStorage;
+
+
+#pragma cyclus exec from cyclus.system import CY_LARGE_DOUBLE, CY_LARGE_INT, CY_NEAR_ZERO
+
+// Use an anonymous namespace to avoid polluting the global namespace with 
+// test-specific code.
+namespace tricycle {
+
+Composition::Ptr tritium() {
+  cyclus::CompMap m;
+  m[10030000] = 0.10;
+  return Composition::CreateFromAtom(m);
+};
+
+Composition::Ptr decayed_tritium() {
+  cyclus::CompMap m;
+  m[10030000] = 0.90;
+  m[20030000] = 0.10;
+  return Composition::CreateFromAtom(m);
+};
+
+std::string common_config =
+    " <incommod>Tritium</incommod>"
+    " <outcommod>Tritium_Out</outcommod>"
+    " <throughput>10</throughput>"
+    " <max_tritium_inventory>10</max_tritium_inventory>";
+
+cyclus::MockSim InitializeSim(std::string config, int simdur) {
+  cyclus::MockSim sim(cyclus::AgentSpec(":tricycle:DecayStorage"), config,
+                      simdur);
+
+  sim.AddRecipe("tritium", tritium());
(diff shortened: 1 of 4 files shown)


src/decay_storage_tests.cc
```diff
@@ -0,0 +1,277 @@
...
+
+  int simdur = 3;
+  cyclus::MockSim sim = InitializeSim(common_config, simdur);
+  // No sink added, so material will not be sold and will accumulate
+
+  int id = sim.Run();
+
+  QueryResult qr_1 = TimeInventoryQuery(sim, "1");
+  double tritium_1 = qr_1.GetVal<double>("TritiumStorage");
+  double helium_1 = qr_1.GetVal<double>("HeliumStorage");
```
gonuke, line 192: I think we expect this to be 0 on time step 1 because none of the tritium that we received will have had a chance to decay.  This would be good to test as a contrast to the `MaterialWithDecay` test below.

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,277 @@
...
+  sim.AddRecipe("decayed_tritium", decayed_tritium());
+  sim.AddSource("Tritium").recipe("decayed_tritium").Finalize();
+
+  int id = sim.Run();
+
+  QueryResult qr = TimeInventoryQuery(sim, "1");
+  double helium_qty = qr.GetVal<double>("HeliumStorage");
+
+  // Should extract helium-3 from the already decayed material
+  EXPECT_LE(0.0, helium_qty);
```
gonuke: I'm surprised that this is true since on time step 1 we will have received this material that contains He-3, but we won't extract it until the `Tick` of the next time step.

gonuke (changes requested):
This looks great now!  (Too bad you had to make a new PR, though...)

I do have one lingering question about the correct behavior of a couple of tests - maybe @dean-krueger can chime in as well.

src/decay_storage_tests.cc
gonuke: Oops - off-by-one - first trade happens on time step 0

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,277 @@
...
+
+  QueryResult qr_2 = TimeInventoryQuery(sim, "2");
+  double tritium_2 = qr_2.GetVal<double>("TritiumStorage");
+  double helium_2 = qr_2.GetVal<double>("HeliumStorage");
+
+  // Verify we have material at time 1
+  EXPECT_LT(0.0, tritium_1);
+
+  // Helium-3 should accumulate as tritium decays
+  EXPECT_LT(helium_1, helium_2);
```
gonuke, line 203: While testing helium_1 < helium_2 implies that all facets of the decay storage decay and helium extraction are working, this additional test adds clarity at little cost.
```suggestion
  EXPECT_LT(0.0, helium_1);
  EXPECT_LT(helium_1, helium_2);
```

src/decay_storage_tests.cc
```diff
@@ -0,0 +1,277 @@
...
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, EnterNotifyPolicySetup) {
+  // Test that EnterNotify sets up buy and sell policies correctly
+
+  EXPECT_NO_THROW(facility->EnterNotify());
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(DecayStorageTest, MaterialWithDecay) {
```
gonuke: This tests seems redundant with `DecayAndExtractHelium` and may not be necessary.

src/decay_storage_tests.cc
gonuke: oops - off-by-one - first trade happens on time step 0 and extraction happens in the Tick of time step 1 before this is recorded

gonuke (changes requested):
One small change to add a test for clarity (even if existing tests demonstrate correctness).

gonuke (approved):
LGTM - sorry for the delay in review

PR #78, opened 2026-05-28
Added fuel out commodities to FPP archetype

trysarahtops:
### Summary of Changes
This PR adds fuel outcommodity as an input for the Fusion Power Plant archetype in tricycle. It allows the output commodity to be set differently than the input commodity.

### DIFF
--- a/src/fusion_power_plant.h
+++ b/src/fusion_power_plant.h
@@ -124,7 +124,7 @@ class FusionPowerPlant : public cyclus::Facility  {
   std::string fuel_incommod;
 
   #pragma cyclus var { "default":"",\
-    "doc": "Fresh fuel commodity", \
+    "doc": "Fuel commodity leaving power plant", \
     "tooltip": "Name of fuel commodity offered", \
     "uilabel": "Fuel output commodity" \
   }
--- a/src/fusion_power_plant.h
+++ b/src/fusion_power_plant.h
@@ -99,7 +99,7 @@ class FusionPowerPlant : public cyclus::Facility  {
   double reserve_inventory;  
 
   #pragma cyclus var { \
-    "doc": "Equilibrium quantity of tritium which is sequestered in the system and no longer accessable", \
+    "doc": "Equilibrium quantity of tritium which is sequestered in the system and no longer accessible", \
     "tooltip": "sequestered tritium equilibrium quantity, should be startup-reserve inventory", \
     "units": "kg", \
     "uilabel": "Equilibrium Quantity of Sequestered Tritium" \
@@ -123,7 +123,14 @@ class FusionPowerPlant : public cyclus::Facility  {
   }
   std::string fuel_incommod;
 
-    #pragma cyclus var { \
+  #pragma cyclus var { "default":"",\
+    "doc": "Fresh fuel commodity", \
+    "tooltip": "Name of fuel commodity offered", \
+    "uilabel": "Fuel output commodity" \
+  }
+  std::string fuel_outcommod;
+
+  #pragma cyclus var { \
     "default": 0.03, \
     "doc": "Fraction of tritium that comes from the (n + Li-7 --> T + He + n) reaction", \
     "tooltip": "Fraction of tritium from Li-7 breeding", \
--- a/src/fusion_power_plant_tests.cc
+++ b/src/fusion_power_plant_tests.cc
@@ -94,13 +94,13 @@ TEST_F(FusionPowerPlantTest, Print) {
   // Test FusionPowerPlant specific aspects of the print method here
 }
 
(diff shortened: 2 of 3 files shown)


src/fusion_power_plant_tests.cc
```diff
@@ -354,14 +355,15 @@ TEST_F(FusionPowerPlantTest, EnterNotifyInvalidFill) {
 
   EXPECT_THROW(int id = sim.Run(), cyclus::KeyError);
 }
-
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// Unfinished
```
gonuke: Is this still true?  Is there more to come?

gonuke (commented):
Thanks @trysarahtops - this looks pretty complete to me, but there is a comment about the tests being unfinished.  Can you confirm?

src/fusion_power_plant.h
```diff
@@ -123,7 +123,14 @@ class FusionPowerPlant : public cyclus::Facility  {
   }
   std::string fuel_incommod;
 
-    #pragma cyclus var { \
+  #pragma cyclus var { "default":"",\
+    "doc": "Fresh fuel commodity", \
```
dean-krueger: Is the phrase "Fresh fuel commodity" true here? Or I guess a better question would be "Is this how we want to represent this in the doc"? Perhaps something a little more verbose would be good? Not entirely sure what that looks like, but this stuck out to me.

src/fusion_power_plant_tests.cc
```diff
@@ -209,12 +209,13 @@ TEST_F(FusionPowerPlantTest, DecayInventoryExtractHelium) {
   double lambda = 2.57208504984001213e-09;
   double t = 2629846;
 
-  double expected_decay = init_quant - init_quant * std::pow(2, -lambda * t);
+  double expected_decay = (init_quant) -
+                          (init_quant) * std::pow(2, -lambda * t);
```
dean-krueger: Are the parentheses around `init_quant` doing anything here/necessary?

src/fusion_power_plant_tests.cc
```diff
@@ -354,14 +355,15 @@ TEST_F(FusionPowerPlantTest, EnterNotifyInvalidFill) {
...
+// Unfinished
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 TEST_F(FusionPowerPlantTest, EnterNotifySellPolicy) {
   // Test sell policy behavior of enter notify.
 
   std::string config = common_config +
                        " <TBR>1.30</TBR> "
-                       " <fuel_incommod>Tritium</fuel_incommod>";
+                       " <fuel_incommod>Tritium</fuel_incommod>"
+                        "<fuel_outcommod>TritiumFuel</fuel_outcommod>";
```
dean-krueger: I think this comes from the naming convention you've used in your simulations, but perhaps something like `TritiumFromFPP` might make more sense here as the `fuel_outcommod`?

src/fusion_power_plant_tests.cc
```diff
@@ -371,7 +373,7 @@ TEST_F(FusionPowerPlantTest, EnterNotifySellPolicy) {
   sim.AddRecipe("enriched_lithium", enriched_lithium());
 
   sim.AddSource("Tritium").capacity(100).recipe("tritium").Finalize();
-  sim.AddSink("Tritium").Finalize();
+  sim.AddSink("TritiumFuel").Finalize();
```
dean-krueger: See above

dean-krueger (commented):
Thanks @trysarahtops , this looks good overall! I saw a few minor things in addition to the unfinished test @gonuke pointed out which I think warrant a quick second pass, but otherwise this makes a lot of sense. Leaving this as a comment since nothing seemed super critical and I don't want to block.

src/fusion_power_plant_tests.cc
trysarahtops: I'll remove this. I was thinking about adding a test for outcommod, but it seems unnecessary to me at this point.

src/fusion_power_plant.h
trysarahtops: Sorry, I used the boilerplate templating and forgot to update.

src/fusion_power_plant_tests.cc
trysarahtops: I forgot to remove them when I was updating the test. I don't think they are necessary.

src/fusion_power_plant_tests.cc
trysarahtops: Sounds good! I think I will change the name here.

gonuke (approved):
Thanks @trysarahtops -this looks good

PR #79, opened 2026-05-30
Flexible fusion plant

ChasingNeutrons:
As part of some work with @amerallaf and @dean-krueger I have developed a new fusion plant facility. This plant is 'flexible' in that it can have an arbitrary number of components between which tritium can migrate over time. This allows relatively easy construction of complicated plant/tritium handling arrangements. The plant also allows plasma disruptions to be represented by specifying a failure frequency; the plant will randomly fail to operate based on this frequency.

Two drawbacks of the plant: first, it does not track other fusion relevant isotopes, such as helium or lithium. Second, it requires an additional dependency, Eigen, in order to evolve the tritium with a matrix exponential approach.

I am quite fresh to Cyclus and very rusty in my C++, so I will be very grateful for the feedback!


src/flexible_fusion_plant.h
```diff
@@ -0,0 +1,362 @@
...
+  virtual void Tick();
+
+  /// The handleTick function specific to the FlexibleFusionPlant.
+  /// @param time the time of the tock
+  virtual void Tock();
+
+  /// A verbose printer for the FlexibleFusionPlant
+  virtual std::string str();
+
+  //State Variables:
```
gonuke, line 146: Normally, all of these state variables will be in the e `private:` section

src/flexible_fusion_plant.h
```diff
@@ -0,0 +1,362 @@
...
+
+  cyclus::toolkit::MatlBuyPolicy fuel_startup_policy;
+  cyclus::toolkit::MatlBuyPolicy fuel_refill_policy;
+
+  cyclus::toolkit::MatlSellPolicy tritium_sell_policy;
+
+  cyclus::toolkit::TotalInvTracker fuel_tracker;
+
+  //This is to correctly instantiate the TotalInvTracker(s)
+  double fuel_limit = 1000.0;
```
gonuke: Since this is only used in once place, you could use a local variable in `EnterNotify()` instead...

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+using cyclus::FixedIntDist;
+using cyclus::IntDistribution;
+using cyclus::KeyError;
+using cyclus::Material;
+using cyclus::toolkit::ResBuf;
+
+namespace tricycle {
+
+const double MW_to_W = 1000000;
+const double lambda_T = std::log(2.0) / (12.32 * 365 * 24 * 3600);
```
gonuke: May want to refer to a `cyclusYear`
```suggestion
const double lambda_T = std::log(2.0) / (12.32 * cyclusYear);
```

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+const double energy_DT = 17.6 * 1.6021766 * std::pow(10,-13);
```
gonuke: We should probably define a variable `MeV_to_J`... and this can just be written:
```suggestion
const double energy_DT = 17.6 * 1.6021766E-13;
```

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+const double mass_tritium = 5.01 * std::pow(10,-27);
```
gonuke: Probably best to get this from the built-in access to PyNE to ensure internal consistency ... and this could be written
```suggestion
const double mass_tritium = 5.01E-27;
```

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+void FlexibleFusionPlant::EnterNotify() {
+  cyclus::Facility::EnterNotify();
+
+  fuel_tracker.Init({&tritium_storage}, fuel_limit);
+  
+  burn_rate = mass_tritium * fusion_power * MW_to_W/ 
+	  (conversion_efficiency * energy_DT);
+  fuel_usage_mass = burn_rate * context()->dt();
+
+  failure_probability = 1.0 - std::exp(-failure_frequency * context()->dt() / 12);
```
gonuke: Magic number 12 - assuming this refers to months in a year... ~~If so, we have a variable for that.~~ I thought we had a variable for that, but we don't - maybe should make an issue.

Is it still valid if the time step length is not 1 month?

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+
+      total += x;
+    }
+
+    // Check total
+    if (total > 1.0) {
+      throw cyclus::ValueError(
+        "Escape fractions must sum to <= 1.");
+    }
+  }
```
gonuke: Some things that can streamline this:
`std::any_of`, `std::all_of`, `std::none_of`: https://en.cppreference.com/cpp/algorithm/all_any_none_of

`std::accumulate` https://en.cppreference.com/cpp/algorithm/accumulate

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+      double fraction = escape_fraction[k];
+
+      A(to, plasma) = fraction * (1 - TBE) * tritium_consumption_rate;
+	
+    }
+  }
+
+  // Add constant removal from storage into plasma
+  int storage = comp_index["storage"];
+  A(storage, plasma) = -tritium_consumption_rate;
```
gonuke, line 168: Is this indexed backwards?  All the other cases are `(to, from)` but this appears to be `(from, to)`

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+
+  // Fill transfer terms
+  if (transfer_rate.size() > 0) {
+    for (int k = 0; k < transfer_rate.size(); k++) {
+
+      if (comp_index.count(transfer_from[k]) == 0) {
+        throw cyclus::ValueError(
+          "Unknown transfer source component: " +
+          transfer_from[k]);
+      }
```
gonuke: Thinking it might make sense to have a small private function:
```
void FlexibleFusionPlant::_require_string(std::map<std::string, int> string_map, std::string required, std::string error_string)
```

(In fact, it probably makes sense to implement this as a utility for use across Cyclus, but maybe useful to you here for now.)

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+void FlexibleFusionPlant::BuildMatrix(double tritium_consumption_rate) {
+
+  // Wipe any existing elements
+  A.setZero();
+      
+  int plasma = comp_index["plasma"];
+
+  // Fill transfer terms
+  if (transfer_rate.size() > 0) {
+    for (int k = 0; k < transfer_rate.size(); k++) {
```
gonuke: I tend to prefer semantically meaningful index variables, e.g. here it could be `flow`

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+  
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+void FlexibleFusionPlant::BuildMatrix(double tritium_consumption_rate) {
+
+  // Wipe any existing elements
+  A.setZero();
+      
+  int plasma = comp_index["plasma"];
```
gonuke: Since `"plasma"` is required and can't be used for a transfer, does it make sense to keep it out of the component list and then automatically add it during initialization?

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+      ->Record();
+}
+
+double FlexibleFusionPlant::SequesteredTritium() {
+  double current_sequestered_tritium = 0.0;
+
+  for (int i = 0; i < components.size(); i++) {
+    
+    if (i == comp_index["plasma"] || tritium_elsewhere[i].empty()
+	|| i == comp_index["storage"]) {continue;}
```
gonuke: Style - indent this further (Google C++ style guide and/or clang-format)

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+  
+  // Otherwise the ResBuf encounters an error when it tries to squash
+  if (excess_tritium > cyclus::eps_rsrc()) {
+    tritium_excess.Push(tritium_storage.Pop(excess_tritium));
+  }
+
+  if (SequesteredTritium() != 0) {
+    fuel_startup_policy.Stop();
+    fuel_refill_policy.Start();
+  }
```
gonuke: Not sure the operational intent here?  Are you suggesting that any amount of sequestered tritium means that you no longer are in startup mode but are in refill mode? Just checking because we've had a lot of discussion about this for the FPP.

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+  // Wipe any existing elements
+  A.setZero();
+      
+  int plasma = comp_index["plasma"];
+
+  // Fill transfer terms
+  if (transfer_rate.size() > 0) {
+    for (int k = 0; k < transfer_rate.size(); k++) {
+
+      if (comp_index.count(transfer_from[k]) == 0) {
```
gonuke: Since you call `BuildMatrix` frequently, it may make sense to perform these tests once during `EnterNotify` or so and not each time you run `BuildMatrix`

src/flexible_fusion_plant.cc
gonuke: If you automatically added it at the end of the list, you could then run future loops to not include the plasma and avoid having to test whether your given row is the plasma or not if you want to ignore plasma anyway.

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+      throw cyclus::ValueError(
+        "Escape fractions must sum to <= 1.");
+    }
+  }
+
+  // Create matrix
+  A = Eigen::MatrixXd::Zero(N, N);
+
+  // Build the matrix
+  BuildMatrix(burn_rate);
```
gonuke: Do you ever use this matrix that you build?  I think you rebuild it before you ever use it?

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,431 @@
...
+    } else if (i == comp_index["storage"]) {
+      tritium_vector(i) = tritium_storage.quantity();
+    } else {
+      tritium_vector(i) = tritium_elsewhere[i].quantity();
+    }
+
+  }
+  
+  // Set the source term
+  BuildMatrix(Q);
```
gonuke: It seems that you will only ever have two versions of this matrix, one during burn and one during a down time.  I think it makes sense to build these each once and then decide which you are using.

gonuke (changes requested):
Thanks for this contribution @ChasingNeutrons - looks interesting.

I only see one algorithmic issue, otherwise just some suggestions to simplify and streamline things.

src/flexible_fusion_plant.cc
ChasingNeutrons: This is a good point. My original intent was to allow partial burns of varying durations, but I backed away from that for now because of the complexity. It shouldn't be too hard to revive later if needed. Meanwhile I will make two matrices.

src/flexible_fusion_plant.cc
ChasingNeutrons: This snippet is actually a hangover from starting with fusion_power_plant. Thanks for raising this as I think it is redundant with the beginning of the function (once ReadyToOperate returns true, the same policy actions occur). As such, I am inclined to remove this.

src/flexible_fusion_plant.cc
ChasingNeutrons: This is not indexed backwards, but I also had to think about this a bit when I first wrote it. The plasma index in the tritium vector is just a constant 1 at all times. As such, this term in the governing differential equation corresponds to a constant removal rate from the storage. This is analogous with the breeder down below, only that it is a constant production rate. I have added some additional comments to hopefully make this more clear.

src/flexible_fusion_plant.cc
ChasingNeutrons: I fear my Fortran is showing... Thanks for that!

src/flexible_fusion_plant.cc
ChasingNeutrons: This was actually an error on my part, so thanks for spotting it. I had previously defined failure_frequency to be per month. However, the context()->dt() is in seconds. I don't quite know how I ended up with the 12 other than indeed thinking months per year was necessary. I have instead redefined failure_frequency to more optimistically be per year, and have used cyclusYear as the conversion for dt, as suggested above.

src/flexible_fusion_plant.cc
ChasingNeutrons: Nice suggestion - this has simplified quite a few bits of the code.

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,450 @@
...
+  
+  // Add the tritium source term
+  // In analogy with the above, this corresponds to a constant production rate
+  // in the breeder.
+  int breeder = comp_index["breeder"];
+  A(breeder, plasma) = TBR * TBE * tritium_consumption_rate;
+
+  // Also add diagonal tritium decay term
+  for (int component = 0; component < components.size(); component++) {
+    A(component, component) -= lambda_T;
```
gonuke: I don't think you need the variable `lambda_T`
```suggestion
    A(component, component) -= pyne::decay_const(tritium);
```

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,450 @@
...
+namespace tricycle {
+
+const int tritium_id = 10030000;
+const cyclus::CompMap T = {{tritium_id, 1}};
+const cyclus::Composition::Ptr tritium_comp = cyclus::Composition::CreateFromAtom(T);
+const double mass_tritium = pyne::atomic_mass(tritium_id) / (1000.0 * pyne::N_A);
+
+const double MW_to_W = 1000000;
+const double MeV_to_J = 1.6021766E-13;
+const double lambda_T = std::log(2.0) / (12.32 * cyclusYear);
```
gonuke: I don't think you need this variable - see below

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,450 @@
...
+    
+    // Check validity of destinations
+    for (int flow = 0; flow < escape_fraction.size(); flow++) {
+
+      _require_string(comp_index, escape_to[flow],
+		      "Unknown escape destination component: " + 
+		      escape_to[flow]);
+    }
+
+  }
```
gonuke: Style preference - it probably makes sense to put this in its own function, e.g. `void FlexibleFusionPlant::ValidateInput()`.  This mostly reduced the cognitive burden of parsing what happens in `EnterNotify()` for long term readability/maintainability.

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,450 @@
...
+  int N = components.size() + 1;
+  
+  // For legibility: plasma is the last element
+  int plasma = N - 1;
+  
+  Eigen::MatrixXd A = Eigen::MatrixXd::Zero(N,N);
+      
+  // Fill transfer terms
+  if (transfer_rate.size() > 0) {
+    for (int flow = 0; flow < transfer_rate.size(); flow++) {
```
gonuke: This conditional is not necessary - if the size is 0 the loop will just do nothing

src/flexible_fusion_plant.cc
```diff
@@ -0,0 +1,450 @@
...
+      A(to, from) += rate;
+
+      // Diagonal loss to other component
+      A(from, from) -= rate;
+
+    }
+  }
+
+  // Fill plasma escape terms
+  if (escape_fraction.size() > 0) {
```
gonuke: ibid - don't need conditional

gonuke (changes requested):
Found a few pending comments that I forgot to share (probably waiting for time to do further review)

PR #81, opened 2026-07-28
Inclusion of startup policy

trysarahtops:
### Summary of Changes
This PR introduces Startup Policy to ensure that a minimum quantity of material is present in order to startup an archetype.

### Design Notes
This policy is initially added to the FusionPowerPlant archetype to test functionality. Updates need to be made to ensure that all unit testing passes. 

### Associated Developers
Claude

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,7 @@
 ### DO NOT DELETE THIS COMMENT: INSERT_ARCHETYPES_HERE ###
 USE_CYCLUS("tricycle" "fusion_power_plant")
 USE_CYCLUS("tricycle" "decay_storage")
+USE_CYCLUS("tricycle" "strt_up_policy")
 INSTALL_CYCLUS_MODULE("tricycle" "")
 
 # install header files
--- a/src/fusion_power_plant.cc
+++ b/src/fusion_power_plant.cc
@@ -1,5 +1,7 @@
 #include "fusion_power_plant.h"
+#include "strt_up_policy.h"
 
+using tricycle::StartupPolicy;
 using cyclus::CompMap;
 using cyclus::Composition;
 using cyclus::DoubleDistribution;
@@ -40,6 +42,11 @@ void FusionPowerPlant::EnterNotify() {
   fuel_usage_mass = (burn_rate * (fusion_power / MW_to_GW) /
                      (kDefaultTimeStepDur * 12) * context()->dt());
   blanket_turnover = blanket_size * blanket_turnover_fraction;
+
+  strt_up_policy.Register(
+      "tritium_storage", &tritium_storage,
+      (reserve_inventory + sequestered_equilibrium) * tritium_startup_fraction);
+  strt_up_policy.Register("blanket_feed", &blanket_feed, blanket_turnover);
 
   // Create the blanket material for use in the core, no idea if this works...
   blanket = Material::Create(this, 0.0, context()->GetRecipe(blanket_inrecipe));
@@ -175,15 +182,14 @@ bool FusionPowerPlant::TritiumStorageClean() {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 bool FusionPowerPlant::ReadyToOperate() {
   // Determine tritium inventory required to operate
-  double required_storage_inventory = SequesteredTritiumGap();
-  if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
-    required_storage_inventory += reserve_inventory;
-    required_storage_inventory *= tritium_startup_fraction;
-  } else {
-    required_storage_inventory += fuel_usage_mass;
+ if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
+    // Initial startup: gated entirely by strt_up_policy.
+    return strt_up_policy.full();
   }
 
-  // check  tritium storage quantity requirement
+  // Steady-state operation: dynamic per-tick requirement, not a fixed
+  // "startup" threshold, so this stays as bespoke logic.
+  double required_storage_inventory = SequesteredTritiumGap() + fuel_usage_mass;
   if (tritium_storage.quantity() < required_storage_inventory ||
       !TritiumStorageClean()) {
     return false;
@@ -193,6 +199,7 @@ bool FusionPowerPlant::ReadyToOperate() {
   }
   return true;
 }
+
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 void FusionPowerPlant::LoadCore() {
--- a/src/fusion_power_plant.h
+++ b/src/fusion_power_plant.h
@@ -4,8 +4,10 @@
 #include <string>
 
 #include "cyclus.h"
+#include "strt_up_policy.h"
 #include "boost/shared_ptr.hpp"
 #include "pyne.h"
+/// #include "strt_up_policy_snippet.cycpp.h"
 
 using cyclus::Material;
 
@@ -36,6 +38,7 @@ namespace tricycle {
 /// This section needs to be filled out once there is some behavior to describe.
 ///
 class FusionPowerPlant : public cyclus::Facility  {
+  friend class FusionPowerPlantTest;
  public:
   /// Constructor for FusionPowerPlant Class
   /// @param ctx the cyclus context for access to simulation-wide parameters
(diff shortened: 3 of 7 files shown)


src/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 ### DO NOT DELETE THIS COMMENT: INSERT_ARCHETYPES_HERE ###
 USE_CYCLUS("tricycle" "fusion_power_plant")
 USE_CYCLUS("tricycle" "decay_storage")
+USE_CYCLUS("tricycle" "strt_up_policy")
```
gonuke: This line should only be necessary for new archtypes and not for new capability that happens to be in a different file.

src/strt_up_policy.h
```diff
@@ -0,0 +1,104 @@
...
+  /// policy to gate startup should register at least one tracker.
+  bool full() const;
+
+  /// Convenience accessor: is a specific tracked inventory (by name) at
+  /// or above its threshold? Returns false if `name` isn't registered.
+  bool full(const std::string& name) const;
+
+  /// Number of trackers currently registered.
+  int n_trackers() const;
+
```
gonuke, line 89: As discussed, we may want an method to update the threshold:
```
void UpdateThreshold(const std::string& name, double new_threshold);
```

src/strt_up_policy.cc
```diff
@@ -0,0 +1,47 @@
+#include "strt_up_policy.h"
+
+namespace tricycle {
+
+
+StartupPolicy::StartupPolicy() {}
+
+StartupPolicy::~StartupPolicy() {}
```
gonuke, line 8: These could be defined as empty methods in the header file.  Since they are empty, it may be better to just not define them at all and rely on default constructor/destructor.

src/strt_up_policy.h
```diff
@@ -0,0 +1,104 @@
...
+ private:
+   struct Tracker {
+    Tracker(std::string name_, ResBuf<Material>* buf_, double threshold_)
+        : name(name_), buf(buf_), threshold(threshold_) {}
+    std::string name;
+    ResBuf<Material>* buf;
+    double threshold;
+  };
+
+  std::vector<Tracker> trackers_;
```
gonuke, line 99: Given the way this data structure is accessed, it may be better to make it a map, with corresponding changes to the various other methods.

```suggestion
   struct Tracker {
    Tracker(ResBuf<Material>* buf_, double threshold_)
        : buf(buf_), threshold(threshold_) {}
    ResBuf<Material>* buf;
    double threshold;
  };

  std::map<std::string, Tracker> trackers_;
```

src/fusion_power_plant.h
```diff
@@ -4,8 +4,10 @@
 #include <string>
 
 #include "cyclus.h"
+#include "strt_up_policy.h"
 #include "boost/shared_ptr.hpp"
 #include "pyne.h"
+/// #include "strt_up_policy_snippet.cycpp.h"
```
gonuke, line 10: remove this commented line

src/fusion_power_plant.h
```diff
@@ -4,8 +4,10 @@
 #include <string>
 
 #include "cyclus.h"
+#include "strt_up_policy.h"
```
gonuke, line 7: In general, we like to include header files in other header files as rarely as possible. In this case, you can "forward declare" the class below:
```
class tricycle::StartupPolicy;
```
and avoid including the header.  This works because nowhere in this header file do you need to know anything about the public interface to `StartupPolicy`, only that the class exists.

gonuke: If you do this inside the `namespace tricycle` then it only needs to be:
```
class StartupPolicy;
```

src/fusion_power_plant.cc
```diff
@@ -1,5 +1,7 @@
 #include "fusion_power_plant.h"
+#include "strt_up_policy.h"
 
+using tricycle::StartupPolicy;
```
gonuke, line 4: I don't think you need this line.  This is only necessary if you want to refer to the `StartupPolicy` class without having to use the full namespace.  In this case, you have two reasons that you don't need this:
1. You never refer to the class, just use the interface to the class
2. You are already in the `tricycle` namespace, so don't need to declare it

src/fusion_power_plant.cc
```diff
@@ -175,15 +182,14 @@ bool FusionPowerPlant::TritiumStorageClean() {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 bool FusionPowerPlant::ReadyToOperate() {
   // Determine tritium inventory required to operate
-  double required_storage_inventory = SequesteredTritiumGap();
-  if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
-    required_storage_inventory += reserve_inventory;
-    required_storage_inventory *= tritium_startup_fraction;
-  } else {
-    required_storage_inventory += fuel_usage_mass;
```
gonuke, line 183: This was setup to determine a new required level of tritium depending on the system's state.  It was always a little flaky and hard to follow, but if you want to use the new `StartupPolicy` for this, you'll probably need to be able to update the threshold using a similar logic.

It's probably best to match this behavior now and consider how we can model this better going forward under a different issue/PR.

src/fusion_power_plant.cc
```diff
@@ -175,15 +182,14 @@ bool FusionPowerPlant::TritiumStorageClean() {
...
+    // Initial startup: gated entirely by strt_up_policy.
+    return strt_up_policy.full();
   }
 
-  // check  tritium storage quantity requirement
+  // Steady-state operation: dynamic per-tick requirement, not a fixed
+  // "startup" threshold, so this stays as bespoke logic.
+  double required_storage_inventory = SequesteredTritiumGap() + fuel_usage_mass;
   if (tritium_storage.quantity() < required_storage_inventory ||
       !TritiumStorageClean()) {
```
gonuke, line 194: If `ReadyToOperate()` is always called immediately following `ExtractHelium()` then this test for `TritiumStorageClean()` may be unnecessary.  It will be interesting to see if we still pass all the tests without it.

src/fusion_power_plant.cc
```diff
@@ -175,15 +182,14 @@ bool FusionPowerPlant::TritiumStorageClean() {
...
+ if (sequestered_tritium->quantity() < cyclus::eps_rsrc()) {
+    // Initial startup: gated entirely by strt_up_policy.
+    return strt_up_policy.full();
   }
 
-  // check  tritium storage quantity requirement
+  // Steady-state operation: dynamic per-tick requirement, not a fixed
+  // "startup" threshold, so this stays as bespoke logic.
+  double required_storage_inventory = SequesteredTritiumGap() + fuel_usage_mass;
   if (tritium_storage.quantity() < required_storage_inventory ||
```
gonuke, line 193: If the test for `TritiumStorageClean()` is confirmed to be redundant/unnecessary, then we can probably replace this with your new capability:

```suggestion
  if (strt_up_policy.full("tritium"))
```

src/fusion_power_plant.cc
gonuke:
```suggestion
  if (BlanketCycleTime() && strt_up_policy.full("blanket_feed")) {
```

src/fusion_power_plant.h
```diff
@@ -257,6 +260,8 @@ class FusionPowerPlant : public cyclus::Facility  {
   cyclus::toolkit::ResBuf<cyclus::Material> helium_excess;
   cyclus::toolkit::ResBuf<cyclus::Material> blanket_feed;
   cyclus::toolkit::ResBuf<cyclus::Material> blanket_waste;
+  cyclus::toolkit::ResBuf<cyclus::Material> feed_inv;
+  cyclus::toolkit::ResBuf<cyclus::Material> topup_inv;
```
gonuke, line 264: You don't appear to use either of these new `ResBuf`s.

gonuke (changes requested):
I've tried to capture the discussion we had today.

The more I look at the details of this, however, the more I see value in some of this being implemented as changes to `cyclus::toolkit::ResBuf`.

If that class had:
* a new data member: `double full_quantity` that was, by default, equal to the `capacity`
* a new function: `bool full()` that returned `quantity >= full_quantity`
* a new function: `void full_quantity(double new_full_quantity)` that could set/update the value of `full_quantity`

Then an archtype could just check whether the `ResBuf` was full.
cyclus/cyclus#1967

src/fusion_power_plant.h
trysarahtops: I tried to forward declare this a few different ways and none of them seemed to work. Is there some special functionality of cyclus that affects forward declaration?

src/strt_up_policy.cc
trysarahtops: I put these in the header file with defaults.

